PR #5, opened 2016-06-16
Add `/utf8` codec

davidar:
The prefix has been chosen to be compatible with the [UTF-8 BOM](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8)


diasdavid:
awesome! LGTM

PR #16, opened 2016-09-25
The unified multicodecs theory

diasdavid:
# The unified multicodecs theory.

> The theory that unites all the self-described multiformats for a beautifully colored 🚲🏚

**tl;dr;** This PR updates the multicodecs table to incorporate all the multiformats binary packed tables (multihash and multiaddr) into one.

With the introduction of [CIDv1](https://github.com/ipfs/specs/issues/130), we needed a way to describe several types of data(multicodec like) in a way that a program could parse(self described), that didn't increase the data size by much, and so, [multicodec-packed](https://github.com/multiformats/multicodec/blob/master/multicodec-packed.md) was born.

One of the requirements for multicodec-packed is that it needed to be as extensible as the normal multicodec, so that applications developers and protocol engineers could add their own multicodec tables (dictionaries) of their own data structures.

Once we added this in, we realized how multihash, multiaddr or even multibase are just users of multicodec-packed with their own custom tables already.

Given this, we found that it is extremely convinient to have all of these tables converge into one, so that we can expand it overtime and avoid code clashes in table extensions.

This PR fixes some spec errors and merges the multiaddr, multihash and multibase tables in a new table with a more clear format.

TODO:
- [x] Solve the clash issue (udp and sha1 share the same code)
- [x] Get the table reviewed by @jbenet 

Future work:
- ~~Bring in all the mimetypes and give them an packed code~~ Seee https://github.com/multiformats/multicodec/pull/16#issuecomment-258977411


README.md
```diff
@@ -115,18 +91,99 @@ An example of a _great_ path name is:
...
+|                               | /sha3-512/      |                         | 0x14        | n/a            |
+|                               | /shake-128/     |                         | 0x18        | n/a            |
+|                               | /shake-256/     |                         | 0x19        | n/a            |
+|                               | /blake2b/       |                         | 0x40        | n/a            |
+|                               | /blake2s/       |                         | 0x41        | n/a            |
+| **Multiaddrs**                                                                                           |
+|                               | /ip4/           |                         | 0x04        | n/a            |
+|                               | /ip6/           |                         | 0x29        | n/a            |
+|                               | /tcp/           |                         | 0x06        | n/a            |
+|                               | /udp/           |                         | 0x11        | n/a            |
```
diasdavid: clash with `sha1`

diasdavid:
# Update

After much discussion (🚲🏚), we've understood that we can improve the clarity of how multicodec, multicodec-packed and multistream are design and their purpose, if we adjust the names of these protocols. In simple terms this means that:

- rename: multicodec-packed -> multicodec
- rename: multicodec -> multistream
- rename: multistream -> multistream-select

This way:

- multicodec becomes the protocol that simply defines `{ codec: code }` pairs (e.g: { ip4: 0x04 })
- multistream becomes the protocol that gets used for describing streams of data in a human readable way (`<varint>/<codec>\n`)
- multistream-select becomes the protocol for handshaking protocols between two endpoints, which uses multistream describe the protocol the endpoints want to speak.

README.md
```diff
@@ -38,67 +38,43 @@ Moreover, this self-description allows straightforward layering of protocols wit
 `multicodec` is a _self-describing multiformat_, it wraps other formats with a tiny bit of self-description:
 
 ```sh
-<multicodec-header><encoded-data>
-# or
-<varint-len><code>\n<encoded-data>
```
diasdavid: ✔️

README.md
```diff
@@ -115,18 +91,99 @@ An example of a _great_ path name is:
...
+| 0x072f626a736f6e2f            | /bjson/         |                         |             | n/a            |
+| 0x082f75626a736f6e2f          | /ubjson/        |                         |             | n/a            |
+| 0x0a2f70726f746f6275662f      | /protobuf/      | Protocol Buffers        |             | n/a            |
+| 0x072f6361706e702f            | /capnp/         | Cap-n-Proto             |             | n/a            |
+| 0x092f666c61746275662f        | /flatbuf/       | FlatBuffers             |             | n/a            |
+| 0x052f726c702f                | /rlp/           | recursive length prefix | 0x60        | n/a            |
+| **Multiformats**                                                                                         |
+| 0x182f6d756c7469636f6465632f  | /multicodec/    |                         | 0x30        | n/a            |
+| 0x162f6d756c7469686173682f    | /multihash/     |                         | 0x31        | n/a            |
+| 0x162f6d756c7469616464722f    | /multiaddr/     |                         | 0x32        | n/a            |
```
diasdavid: ✔️

README.md
```diff
@@ -21,136 +20,150 @@
...
+| mkv             |                         |             |
+|                                                         |
+| **Blockchain formats**                                  |
+|                                                         |
+| **VCS formats**                                         |
+|                                                         |
+| **IPLD formats**                                        |
+| dag-pb          | MerkleDAG protobuf      |             |
+| dag-cbor        | MerkleDAG cbor          |             |
+| eth-rlp         | Ethereum Block RLP      |             |
```
jbenet: add: 
- git
- bitcoin
- stellar

diasdavid:
@jbenet I'm making js-cid complete (as in, being able to handle bs58(multihash), raw multihash, cid, cidStr and cid construction by parts (version, codec, multihash)) and I found myself in a position of 'luck', because neighter CID v0 or CID v1 clash with the multihash table.

For now it is ok, because multihash only starts at 0x11, which gives us 14 more CID versions (if we ever need to change it). However, a thought crossed my mind that CID code should be governed as a multicodec, so that we can then update the CID table, add another version that doesn't represent an incremental update to the previous number, so that all the parsers implemented meanwhile do not break.

README.md
```diff
@@ -21,136 +20,157 @@
...
+mkv,                ,                         0x
+
+blockchain formats
+
+VCS formats
+
+IPLD formats
+dag-pb,             MerkleDAG protobuf,       0x70
+dag-cbor,           MerkleDAG cbor,           0x71
+eth-block,          Ethereum Block (RLP),     0x90
```
whyrusleeping: I don't think these should be above `0x80`.

README.md
```diff
@@ -21,136 +20,157 @@
...
+
+Multicodecuses "protocol tables" to agree upon the mapping from one multicodec code (a single varint). These tables can be application specific, though -- like [with](https://github.com/multiformats/multihash) [other](https://github.com/multiformats/multibase) [multiformats](https://github.com/multiformats/multiaddr) -- we will keep a globally agreed upon table with common protocols and formats.
+
+## Multicodec table
+
+```csv
+codec,              description,              code
+
+miscelaneous
+bin,                raw binary,               0x00
```
whyrusleeping: I'm not sure i like zero being 'binary'. Zero is a default value in a lot of cases and this makes it difficult to tell between 'unknown/invalid' and 'binary'

whyrusleeping:
@diasdavid can we sit down today and finish this? Its blocking my work on ipld

README.md
Kubuxu: Binary could be 0x55 which is `0b10101010` and is also value of Ethernet frame preamble.

diasdavid:
@whyrusleeping sure, why is it blocking though? The IPLD formats have codes.

whyrusleeping:
I'm taking `0x22` for murmur3

wanderer:
why not reserve the top level table only for the different encoding types? so something like
```
0x00 an encoding table
0x01 bases encodings
0x02 serialization formats
0x03 multicodec
0x04 multihash
0x05 multiaddr
0x06 multibase
0x07 multihashes
0x08 multiaddrs
0x09 archiving formats
0x10 IPLD formats
....
ect
```

0x00 would be specail. What would follow is would be a custom encoding table which would map strings to their binary representation.  This would allow apps to use dynamically create new tables.

wanderer:
> own multicodec tables (dictionaries) of their own data structures.

@diasdavid is a proposed format for the dictionaries?

whyrusleeping:
@diasdavid i'm also going to move forward with using `0x55` as binary.

README.md
whyrusleeping: Because that means they take up two bytes when encoded. Anything that is going to be commonly used we would really love to have take only a single byte.

whyrusleeping: :cry:

kumavis:
@jbenet @diasdavid I have requested codes for the keccak-* family
https://github.com/multiformats/multicodec/pull/19

README.md
kumavis: I noticed `sctp` is also above 0x80 (0x84)

README.md
```diff
@@ -21,136 +20,157 @@
...
+> **Q. Why?**
 
-The one problem is figuring out what we're speaking. Humans are pretty smart, we pick up all sorts of languages over time. And we can always resort to pointing and grunting (the ascii of humanity).
+Because [multistream](https://github.com/multiformats/multistream) is too long for identifiers. We needed something shorter.
 
-Programs have a harder time. You can't keep piping json into a protobuf decoder and hope they align. So we have to help them out a bit. That's what multicodec is for.
+> **Q. Why varints?**
 
-> **Q. Why "codec" and not "encoder" and "decoder"?**
+So that we have no limitation on protocols. Implementation note: you do not need to implement varints until the standard multicodec table has more than 127 functions.
```
Kubuxu: That is the case already.

README.md
```diff
@@ -21,136 +20,157 @@
...
+base2,              binary (0 and 1),         0x00
+base8,              octal,                    0x07
+base10,             decimal,                  0x09
+base16,             hexadecimal,              0x
+base32,             rfc4648,                  0x
+base32hex,          rfc4648,                  0x
+base58flickr,       base58 flicker,           0x
+base58btc,          base58 bitcoin,           0x
+base64,             rfc4648,                  0x
+base64url,          rfc4648,                  0x
```
Kubuxu: The codes of multibase should be the same as https://github.com/multiformats/multibase/blob/master/multibase.csv

README.md
```diff
@@ -21,136 +20,157 @@
...
+sha2-512,           ,                         0x13
+sha3-224,           ,                         0x17
+sha3-256,           ,                         0x16
+sha3-384,           ,                         0x15
+sha3-512,           ,                         0x14
+shake-128,          ,                         0x18
+shake-256,          ,                         0x19
+blake2b,            ,                         0x40
+blake2s,            ,                         0x41
+reserved for apps,  appl specific range,      0x4000-0x40f
```
Kubuxu: There is 0 missing in the end of the range hex number.

Kubuxu:
This should be finalized, we are currently working on a code that uses this spec but the spec if far from done.

README.md
diasdavid: Transport multicodecs are 'more okay' because they get transferred less, in the other hand, format multicodecs get transferred every time a block is transferred, so a byte actually means a lot.

README.md
diasdavid: True

README.md
diasdavid: Not possible, the multibase.csv are chars available in the encodec space of the base, not hex values.

diasdavid:
What we need to do to finalise:

- [ ] Decide if we move binary/base2 to [0x55](https://github.com/multiformats/multicodec/pull/16#pullrequestreview-3905986)
- [ ] Stamp the remaining multicodec values

Things to discuss:

- [ ] Are we happy with having formats taking 2 bytes? (Probably not, is it something that people are willing to accept? Can we reduce it?)

README.md
Kubuxu: Right.

whyrusleeping:
maybe the right way to do this might actually to be having a separate table for each 'set' of codes, (one for hashes, one for formats, etc). and then have codes for each of those tables so that the 'two byte' representation can be shortened to a single byte when we have context, and left as the two bytes form when we don't

whyrusleeping:
claiming `0xc0` and `0xc1` for zcash blocks and transactions respectively. I may need more in case they have other formats to deal with around all the crypto

kumavis:
@whyrusleeping so we dont lose track, I reccomend creating a PR into this PR, like I did here https://github.com/multiformats/multicodec/pull/19

README.md
```diff
@@ -21,136 +20,157 @@
 
 ## Motivation
 
-Multicodecs are self-describing protocol/encoding streams. (Note that a file is a stream). It's designed to address the perennial problem:
+[Multistreams](https://github.com/multiformats/multistream) are self-describing protocol/encoding streams. Multicodec uses an agreed-upon "protocol table". It is designed for use in short strings, such as keys or identifiers (i.e [CID](https://github.com/ipld/cid)).
```
Kubuxu: Do all protocols need to be in 'agreed upon "protocol table"' or only those using packed format?

README.md
```diff
@@ -21,136 +20,157 @@
...
 
-> I have a bitstring, what codec is the data coded with!?
+## Protocol Description - How does the protocol work?
 
-Instead of arguing about which data serialization library is the best, let's just pick the simplest one now, and build _upgradability_ into the system. Choices are never _forever_. Eventually all systems are changed. So, embrace this fact of reality, and build change into your system now.
+`multicodec` is a _self-describing multiformat_, it wraps other formats with a tiny bit of self-description. A multicodec identifier is both a varint and the code identifying the following data, this means that the most significant bit of every multicodec code is reserved to signal the continuation.
```
Kubuxu: It describes only the packed format, I think we aren't yet removing the path format, are we?

Kubuxu:
Ok, so we are not moving away from path codec as for https://github.com/multiformats/multicodec/issues/21#issuecomment-257171451

But then this PR removes all mention about path based codec descriptor.

jbenet:
@diasdavid can you address @Kubuxu's comments about the "path vs code" question? i.e. where do the multicodec paths fit? (we discussed it in shanghai but don't recall the nuances/conclusions)

diasdavid:
@jbenet @Kubuxu The path scheme still exists under multistream, which is what we have been doing all along each time we want to describe a stream of data. See: https://github.com/multiformats/multistream, specially https://github.com/multiformats/multistream#the-protocol-path.

Kubuxu:
My point is this spec should also mention them, ss they are multicodecs too.

diasdavid:
@Kubuxu here:  https://github.com/multiformats/multicodec/pull/16/commits/fa3197fbf890ac28b4b5f3fdde6cbcb76d7e66f1

diasdavid:
> As I go through this PR review, I'll try to address things in different comments so they are easy to be linked, bare with me as I tackle one by one

### Use a table of tables

@wanderer and @whyrusleeping proposed to use a table of tables (https://github.com/multiformats/multicodec/pull/16#issuecomment-255787115 and https://github.com/multiformats/multicodec/pull/16#issuecomment-253351745). Feedback:

- This could have been useful at the beginning and would reduce possible conflicts, (like sha1 and udp).
- It would still require some table code sync, so that no two tables have the same code.
- **Making this move would break all of the multihashes issued so far**, which is a dealbreaker for this proposal

diasdavid:
### multicodec path

I've added a reference to multistream (aka multicodec paths)) here:  https://github.com/multiformats/multicodec/pull/16/commits/fa3197fbf890ac28b4b5f3fdde6cbcb76d7e66f1

//cc @Kubuxu

diasdavid:
### mimetypes

We've considered several times to source the mimetypes table to complete the multicodec table, however, in a second thought and now that we have expansibility in mind, digesting the whole mimetype table might just add a ton of noise (i.e vnd.hzn-3d-crossword). I believe that we expand as we go, keeping the reference to the mimetypes table around, so that each type we cross a data type that is already represented there, we use the same name.

For reference - http://www.iana.org/assignments/media-types/media-types.xhtml

diasdavid:
### binary has been migrated from 0x00 to 0x55

As proposed by @Kubuxu and @whyrusleeping and later set in stone with the merge of IPLD PR in go-ipfs, the code for binary moved from 0x00 to 0x55

README.md
diasdavid: @whyrusleeping is this still a concern? I believe it stopped being as soon you merged IPLD in go-ipfs into master, correct?

diasdavid:
# sha1 and udp clash solution

I've added an extra 0 byte to `udp` multicodec, so that it doesn't colid with `sha1` anymore. It is more convenient to change `udp` than `sha1` as changing `sha1` would mean that all the multihashes that use it. Adding an extra 0 byte is enough to differentiate it, but keeping the value `11` on the code.

diasdavid:
# Stamp remaining multicodecs

I'm feeling resistant to stamp the remaining multicodecs only by following rules of thumb. Is there any reasoning we can use to pick the remaining unspecified codes? **This is the last item in order to merge this PR**

Kubuxu:
I think that specifying that `0x0040` and `0x40` are something different is really bad idea. This means that we can't use normal varint libraries but we will need to craft our own, our spec doesn't specify it and then it isn't really and integer, it is variable length bytestream where highest bit says "there is data still coming".

I asked the question about this quite a time ago: https://github.com/multiformats/unsigned-varint/issues/5
but it was about wether we should canonize such miss-formatted varints before storing them to for example CBOR notation as if we don't do that same object has infinitely many binary CIDs.

Summing up, I think we shouldn't abuse varints in this way, we should specify that the number is prefixed with infinitely many leading zeros and that the canonical form is when there is minimum number of bytes taken to express the number. Then before crafting IPLD objects we would canonize such miss-formatted CIDs which should be fast and easy operation.

jbenet:
@kubuxu can you follow up in https://github.com/multiformats/unsigned-varint/issues/5, or another issue in that repo, with examples? i'm not understanding what you're getting at. Either way, this spec uses https://github.com/multiformats/unsigned-varint so whatever _that_ does, this will do too.  let's resolve it there.

jbenet:
@diasdavid 

> Is there any reasoning we can use to pick the remaining unspecified codes? 

I don't think there's a nice function for assigning these numbers. unfortunately it's a nasty problem, because there's a bunch of "domain specific" functions that may make sense-- but don't make sense throughout.

A function i can think of that works uniformly well (even if not very well) is "first come first assign". meaning, a queue, assign them as they're requested. this is certainly going to make people unhappy about having to use 2 bytes for a commonly used function, but it's unavoidable anyway given there's definitely more than 127 common things users will care about...

So, no. i can't think of a nice way that is not "domain specific and non-universal" or "rules of thumb"  :(  --- but there may be one if people want to keep thinking :)

jbenet:
@diasdavid what is the diff here? what protocols changed value in the table? (i.e. what will break, if anything?)

README.md
jbenet: IIRC we were removing multibases from multicodec packed. a i misremembering, @diasdavid ?

README.md
jbenet: and, there are simply more than 127 things we "really care about". so 2 bytes is in the long run unavoidable.

README.md
jbenet: yes, we probably need to remove this, and add multicodec impls to all the use-case specific multiformats IF their table requires it.

README.md
```diff
@@ -21,136 +20,162 @@
...
+IPLD formats
+dag-pb,             MerkleDAG protobuf,       0x70
+dag-cbor,           MerkleDAG cbor,           0x71
+eth-block,          Ethereum Block (RLP),     0x90
+eth-tx,             Ethereum Tx (RLP),        0x91
+bitcoin-block,      Bitcoin Block,            0xb0
+bitcoin-tx,         Bitcoin Tx,               0xb1
+stellar-block,      Stellar Block,            0xd0
+stellar-tx,         Stellar Tx,               0xd1
 ```
```
jbenet: this table should move to a CSV, and the readme should point to it or embed it.

jbenet:
curious why "binary has been migrated from 0x00 to 0x55" -- what was the reasoning?

Kubuxu:
@jbenet https://github.com/multiformats/multicodec/pull/16#discussion_r83039540

but there was discussion elsewhere on it too.

diasdavid:
> @diasdavid what is the diff here? what protocols changed value in the table? (i.e. what will break, if anything? I see UDP, is that the only one)

Yes, that one and the binary which moved to 0x55 as @Kubuxu mentioned and got stamped in go-ipfs.

> "first come first assign". 

Sounds good to me, will add a note to that in the readme.

diasdavid:
# All right, this looks ready to merge :) 

I'll give it a day and merge it Monday (tomorrow)

diasdavid:
All right, I gave it 3 days, going for the merge 🎉🎉

bitbotch (commented):
You can’t get me

PR #19, opened 2016-10-20
Add keccak-*

kumavis:
Keccak-256 is required for ethereum canonical hash support.

Keccak and SHA3 are related. SHA3 is built on top of Keccak.
see https://en.wikipedia.org/wiki/SHA-3#Instances for how SHA3 parameterizes Keccak.

```
keccak-224 0x1A
keccak-256 0x1B
keccak-384 0x1C
keccak-512 0x1D
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,11 @@ sha3-384,           ,                         0x15
 sha3-512,           ,                         0x14
 shake-128,          ,                         0x18
 shake-256,          ,                         0x19
+keccak-256,         ,                         0x1A
+keccak-448,         ,                         0x1B
+keccak-512,         ,                         0x1C
+keccak-768,         ,                         0x1D
+keccak-1024,        ,                         0x1E
 blake2b,            ,                         0x40
 blake2s,            ,                         0x41
 reserved for apps,  appl specific range,      0x4000-0x40f
--- a/README.md
+++ b/README.md
@@ -98,11 +98,10 @@ sha3-384,           ,                         0x15
 sha3-512,           ,                         0x14
 shake-128,          ,                         0x18
 shake-256,          ,                         0x19
-keccak-256,         ,                         0x1A
-keccak-448,         ,                         0x1B
-keccak-512,         ,                         0x1C
-keccak-768,         ,                         0x1D
-keccak-1024,        ,                         0x1E
+keccak-224,         ,                         0x1A
+keccak-256,         ,                         0x1B
+keccak-384,         ,                         0x1C
+keccak-512,         ,                         0x1D
 blake2b,            ,                         0x40
 blake2s,            ,                         0x41
 reserved for apps,  appl specific range,      0x4000-0x40f


kumavis:
@jbenet this PR is targeting the standing "The unified multicodecs theory" PR's branch (multiformats:spec/update), so you can merge it into that PR.

axic:
The instances should be the same as for SHA3 (224, 256, 384, 512).

Kubuxu:
Not really, numbers next to keccak describe internal core size. The length of hash defines output length.

See: https://en.wikipedia.org/wiki/SHA-3#Instances

axic:
@Kubuxu it seems a bit backwards using the internal state size in your naming as opposed to the output size, which what everyone refers to usually?

axic:
Even `sha3` is referred to with its output size:
```diff
 sha3-224,           ,                         0x17
 sha3-256,           ,                         0x16
 sha3-384,           ,                         0x15
 sha3-512,           ,                         0x14
 shake-128,          ,                         0x18
 shake-256,          ,                         0x19
+keccak-256,         ,                         0x1A
+keccak-448,         ,                         0x1B
+keccak-512,         ,                         0x1C
+keccak-768,         ,                         0x1D
+keccak-1024,        ,                         0x1E
```

Kubuxu:
keccak-X can have arbitrary length, sha-X has constant length.

The closest function we have in set currently is shake-X where X is number of bits defining overall security, in which case keccak's numbers should be divided by two.

Kubuxu:
But from the other side, it is not how the function is defined.

kumavis:
Thanks @axic @Kubuxu 

Names are now:
```
keccak-224,         ,                         0x1A
keccak-256,         ,                         0x1B
```

kumavis:
re-pinging @jbenet 😸  https://github.com/multiformats/multicodec/pull/19#issuecomment-255801510

jbenet:
hey-- sorry for delay. this LGTM, but i haven't looked at it deeply enough to know if there's any conflict or issue. +1 to keccak-* in general. @diasdavid is the one controlling the multicodec table, he's the best one to merge this.

diasdavid:
LGTM. Going to merge this one and add a note in the table that the numbers represent the core size. Will also open issue to reference the need for them to be implemented in multihash

jbenet:
On the keccak naming-- i think @Kubuxu is right here-- we should name it along what the creators name them (which i believe is the internal state size). Multihash already takes care of handling output size, so that shouldn't be too much of a concern. The bigger thing with the fn code is making sure people are super clear on _which_ function they're using (internal size)

jbenet:
From: http://keccak.noekeon.org/specs_summary.html

> Keccak is a family of hash functions that is based on the sponge construction, and hence is a sponge function family. In Keccak, the underlying function is a permutation chosen in a set of seven Keccak-f permutations, denoted Keccak-f[b], where b ∈ {25, 50, 100, 200, 400, 800, 1600} is the width of the permutation. The width of the permutation is also the width of the state in the sponge construction.

axic:
I think none of us were entirely correct. The state width is defined as `r + c`:
> We obtain the Keccak[r,c] sponge function, with parameters capacity c and bitrate r, if we apply the sponge construction to Keccak-f[r+c] and by applying a specific padding to the message input.

The previous naming proposal used `keccak-<c>`. Naming it with `keccak-<b>` seems like a reasonable idea.

The reference implemention has these combinations:
```
        case 0: // Default parameters, arbitrary length output
            InitSponge((spongeState*)state, 1024, 576);
            break;
        case 224:
            InitSponge((spongeState*)state, 1152, 448);
            break;
        case 256:
            InitSponge((spongeState*)state, 1088, 512);
            break;
        case 384:
            InitSponge((spongeState*)state, 832, 768);
            break;
        case 512:
            InitSponge((spongeState*)state, 576, 1024);
            break;
```

This matches the current naming merged by @jbenet.

kumavis:
@axic thank you for the very clear explanation!

PR #23, opened 2016-10-28
add in some more codes

whyrusleeping:
added some more codes, cc @diasdavid


README.md
```diff
@@ -56,7 +56,7 @@ Multicodecuses "protocol tables" to agree upon the mapping from one multicodec c
 codec,              description,              code
 
 miscelaneous
-bin,                raw binary,               0x00
```
diasdavid: @jbenet 👍 or 👎 ?

README.md
```diff
@@ -98,8 +98,10 @@ sha3-384,           ,                         0x15
 sha3-512,           ,                         0x14
 shake-128,          ,                         0x18
 shake-256,          ,                         0x19
+murmur3             ,                         0x22
 blake2b,            ,                         0x40
 blake2s,            ,                         0x41
+dbl-sha2-256        ,                         0x56
```
diasdavid: this is `double` sha2, right?

README.md
```diff
@@ -140,6 +142,8 @@ eth-block,          Ethereum Block (RLP),     0x90
 eth-tx,             Ethereum Tx (RLP),        0x91
 bitcoin-block,      Bitcoin Block,            0xb0
 bitcoin-tx,         Bitcoin Tx,               0xb1
+zcash-block,        Zcash Block,              0xc0
+zcash-tx,           Zcash Tx,                 0xc1
```
diasdavid: 👌🏾

README.md
diasdavid: Have you considered if we can have a way to make this more generic? Reserving a code for a multiple of a hash function seems odd (it is an easy way to throw a lot of codes.

I'm was wondering if we should make this broader and have byte in multihash (specifically multihash) that says how many times it got used, but then again, that would break all the current hashes :( So maybe we have to make this exceptions afterall

PR #24, opened 2016-11-22
add in some more codes

whyrusleeping:


table.csv
```diff
@@ -35,6 +35,7 @@ multihashes,,
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
+dbl-sha2-256,       ,                         0x56
```
lgierth: does this mean double sha2-256? might make sense to also add double sha2-512, cjdns uses that for ipv6 addresses (but on the other hand it truncates to 16 bytes).

diasdavid: I made a comment here with some thoughts on how we can adopt this usages. 

https://github.com/multiformats/multicodec/pull/23#discussion_r86863096

One could argue that a "n times hash function" is a whole new function or we just should have a way to tell the parser that it has to do several rounds.

lgierth: > One could argue that a "n times hash function" is a whole new function

Do you mean a new function on multihash's API, or a new hash function? I think for multihash it's simplest to consider it a new hash function. Apply-n-times is far from the only variation in usage of a hash function, so that might become very complicated very quickly.

whyrusleeping: Yeah, adding a notion of 'rounds' to multihash really opens the door to a lot of complexity. You could hash, truncate, hash, or hash, append salt, hash, etc

whyrusleeping:
@lgierth @diasdavid can we merge this?

table.csv
diasdavid: Yeah, nevermind what I was proposing then, I'm just afraid that it might get out of control.

PR #27, opened 2016-12-23
Add eth-objects

kumavis:
Here are the Ethereum objects as I have mapped them so far. While there are only two underlying structures (RLP and Eth-(Secure-)Trie), they interpret the format of the leaf values based on context, which is lost when resolving at the IPLD level. I sketched out some versions that used fewer IPLD formats by employing some cleverness in the path naming conventions. (see here https://github.com/MetaMask/metamask-plugin/issues/719#issuecomment-267528294). The consensus was to be explicit instead of clever, thusly the list of objects as I've proposed.

I'd additionally like to reserve 0x99-0x9f for additional Ethereum objects I may have missed or may be introduced when Ethereum milestone Metropolis is released.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -171,11 +171,20 @@ mp4,                ,                         0x
 mkv,                ,                         0x
 
 IPLD formats,,
-dag-pb,             MerkleDAG protobuf,       0x70
-dag-cbor,           MerkleDAG cbor,           0x71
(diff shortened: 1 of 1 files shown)


diasdavid:
Going to go ahead and merge this, consider `0x99-0x9f` reserved :)

PR #30, opened 2017-01-05
Add identity hash fn

kumavis:
As discussed here https://github.com/multiformats/multihash/issues/13

We should confirm this is the name we want


kumavis:
The code used here was based off of a misconception. We should still add the identity hash fn but need to determine a name and code.
https://github.com/multiformats/multihash/issues/13#issuecomment-270858821

PR #33, opened 2017-02-05
Add multihash for libp2p-crypto public key

mildred:
Related to ipfs/go-cid#15, I didn't realized that the codes were defined here.


mildred:
I am not going to use this, so except a good idea, for the future, this PR is not useful. For the reference, here is what I suggested:

```
+libp2p,,
+libp2p-pubkey-hash,   Multihash of libp2p-crypto encapsulated public key, 0x0210
```

PR #34, opened 2017-02-05
Add mutable record

mildred:
Allow identifying mutable objects given the hash of an immutable
record definition. To avoid using too much codes, the record definition could be defined to require a multicodec header.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -177,6 +177,7 @@ mkv,                ,                         0x
 IPLD formats,,
 dag-pb,               MerkleDAG protobuf,                               0x70
 dag-cbor,             MerkleDAG cbor,                                   0x71
+iprs,                 Latest version of mutable record identified,      0x78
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -177,6 +177,7 @@ mkv,                ,                         0x
 IPLD formats,,
 dag-pb,               MerkleDAG protobuf,                               0x70
 dag-cbor,             MerkleDAG cbor,                                   0x71
+iprs,                 Latest version of mutable record identified,      0x78
```
diasdavid: As I brought up in https://github.com/multiformats/multicodec/pull/37#issuecomment-285048257

IPRS is just data, not a format, it is not required to create a format for something that doesn't have a different serialization format nor has a specific way to resolve through. I understand that the documentation could be more clear on this.

PR #35, opened 2017-02-14
Add BitTorrent info metadata format as 0x7b

jeremyBanks:
Adds BitTorrent [info metadata](http://www.bittorrent.org/beps/bep_0003.html#info-dictionary) format as `0x7b`. (Mnemonic: `B7` for <b>B</b>it<b>T</b>orrent, but transposed so the value remains one byte after [unsigned-varint](https://github.com/multiformats/unsigned-varint) encoding.) Also adds [BitTorrent's bencoding](http://www.bittorrent.org/beps/bep_0003.html#bencoding) under serialization formats, but leaves it unassigned for now.

I'm working on some software that includes a web gateway for distributed content. It will initially only serve content from BitTorrent torrents, but I'd like to eventually include interlinked IPFS content, so I intend to move to a path scheme that's as similar as possible (ideally the same/compatible) as that used by IPFS/IPLD/CIDv1.

As a first step I believe I need a multicodec assignment for the BitTorrent format, hence this PR.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -26,6 +26,7 @@ flatbuf,            FlatBuffers,              0x
 rlp,                recursive length prefix,  0x60
 msgpack,            MessagePack,              0x
 binc,               Binc,                     0x
+bencoding,          Bencoding,                0x
 
 multiformats,,
 multicodec,         ,                         0x30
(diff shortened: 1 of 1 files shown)


diasdavid:
@jeremyBanks I just started working on adding Torrents support to js-ipfs, track the progress here https://github.com/ipfs/js-ipfs/issues/779, you are welcome to join this endeavour if you are interested :) 

Going to complete with this PR with the torrent-file multicodec and add one code for bencode :)

diasdavid:
Going to merge this into a branch of the repo, so I can add stuff 
image

PR #37, opened 2017-02-25
Add DHT record code

mildred:
As discussed in https://github.com/libp2p/go-libp2p-kad-dht/issues/50#issuecomment-282146921, I want to reference IPFS DHT records using CID. This PR allocates a code for that.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -178,6 +178,10 @@ IPLD formats,,
 dag-pb,               MerkleDAG protobuf,                               0x70
 dag-cbor,             MerkleDAG cbor,                                   0x71
 
+IPFS,,
+ipfs-kad-dht-record,  Key of an IPFS record in the DHT,                 0x0220
+
+Blockchain,,
 eth-block,            Ethereum Block (RLP),                             0x90
 eth-block-list,       Ethereum Block List (RLP),                        0x91
 eth-tx-trie,          Ethereum Transaction Trie (Eth-Trie),             0x92


hsanjuan:
If I follow your comments, this is the code you propose for a custom record format which is referenced from the DHT?

I think it would be good if you accompanied this with an actual implementation of the codec for that format so that we can assure that codes in the table have real usage, but I'm not sure about the policy. @Kubuxu @whyrusleeping can you help?

mildred:
I have working code here : https://github.com/mildred/ipfs-objects/

What it does is advertise a CID for a DHT record. The lookup process for those records is a little different. Instead of looking it up directly in the DHT, this CID is looked up instead, then on the resulting list of peers, a direct record query is isued, only to the peers advertising for that record through the CID.

This allows to advertise records that are not understandable my most peers, because when putting a record in the DHT, if the peer that is designed to store it doesn't understand its meaning and how to validate it, it will refuse to store it.

This method allows only peers that understand the record to actually host it.

Here is the actual function that takes a record key and convert it to a CID: https://github.com/mildred/ipfs-objects/blob/master/src/ipobj/interfaces.go#L23-L25

diasdavid:
DHT Records should follow [IPRS (or an updated version of that spec)](https://github.com/ipfs/specs/tree/master/iprs), they should be standard dag-pb or dag-cbor objects. I'm not seeing a reason to call a record another format.

mildred:
There are many reasons not to follow this spec (simplicity, application specific) but the most important thing for me seems to be that IPRS assumes that every node of the network agrees on predefined validity schemes, and to use a validity scheme, all the network must recognize it.

The [system I'm creating](https://github.com/mildred/ipfs-objects/) doesn't use a validity scheme that is recognized by the whole network. As such, the record cannot be present on just any node of the DHT and there is a need to advertise where this record can be found. To locate something on peers of the DHT, the thing must be identified by a CID. As such, I needed to allocate a code to locate DHT records.

If it is not desirable to have a code allocated for this specifically, I suggest to allocate codes for application specific usages.

PR #41, opened 2017-03-05
add MD4 and MD5 hashes

celeduc:
Resolve [multihash#71](https://github.com/multiformats/multihash/issues/71) by adding MD4 and MD5 hashes.  See [multihash#76](https://github.com/multiformats/multihash/pull/76).

cc @lgierth @diasdavid @whyrusleeping

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -35,8 +35,8 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
-md4,                ,                         0x04
-md5,                ,                         0x05
+md4,                ,                         0xd3
+md5,                ,                         0xd4
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
--- a/table.csv
+++ b/table.csv
@@ -34,6 +34,8 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
+md4,                ,                         0x04
+md5,                ,                         0x05
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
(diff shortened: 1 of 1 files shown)


Kubuxu:
What do you think about moving them into two byte range? They should not be used with content addressing where space is the key.

cc @whyrusleeping @jbenet @dignifiedquire

celeduc:
If there is a proposal for exactly where to put these in the two-byte range, then I'm all for it, and I'll make all of the changes necessary.

Also, I'll remove the file from [multihash](https://github.com/multiformats/multihash/blob/master/hashtable.csv) where these are defined, since there should be only one place where these are maintained.

diasdavid:
> They should not be used with content addressing where space is the key.

@Kubuxu I'm not sure if I understand what you meant here

Kubuxu:
With content addressing you use multihash repeated many times which is linked and so on. This means that one additional byte makes a difference, especially with transfers and so on. IMO as MD5 and MD4 are depracated, it makes sense to save some space in one byte range (0-127) and move it to two byte range.

celeduc:
@Kubuxu Could you please suggest values that meet with your requirements so this can be finished?

Kubuxu:
How about `0xd3` and `0xd4`,

celeduc:
Changed as per @Kubuxu above.

table.csv
```diff
@@ -35,6 +35,8 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
+md4,                ,                         0xd3
+md5,                ,                         0xd4
```
diasdavid: Well, if we are going to arbitrarily pick a number, better make it `0xd4` and `0xd5`. Sounds good @Kubuxu ?

Kubuxu: SGWM

celeduc: Done and done.

celeduc:
Ready for merge!

PR #45, opened 2017-03-22
reserving murmur3 32 bit

pgte:


Kubuxu:
Murmur is currently a bit of a mess as go-ipfs HAMT sharding reports 0x22 for 64bit murmur (which isn't really standardised).

lgierth:
> Murmur is currently a bit of a mess as go-ipfs HAMT sharding reports 0x22 for 64bit murmur (which isn't really standardised).

Let's do that right now? If there's already a code in official-ish use, it should be reflected in the table.

Kubuxu:
What I mean is that Murmur64 isn't really standardised. Also murmur on x86 and x64 has different values.

vmx:
I've tried to rebase it through the GiHub UI, but totally screwed up the commit. Hence I decided to create a new PR: https://github.com/multiformats/multicodec/pull/101.

PR #46, opened 2017-04-22
Add ed25519 public key codec

JustinDrake:
See https://github.com/ipfs/notes/issues/241#issuecomment-296392217

I suggest using the code `0xed` so that's it's visually obvious and easy to remember it's associated with ed25519.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -200,3 +200,4 @@ stellar-tx,           Stellar Tx,                                       0xd1
 
 torrent-info,         Torrent file info field (bencoded),               0x7b
 torrent-file,         Torrent file (bencoded),                          0x7c
+ed25519,              Ed25519 public key,                               0xed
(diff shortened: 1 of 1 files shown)


JustinDrake:
I'm a little bit confused as to whether or not this is a valid varint codec.

Kubuxu:
@JustinDrake values in this table are not varint encoded. Varint encoded values might look different.

JustinDrake:
@Kubuxu I see. Then maybe `0xed` is not the wisest of choices because when varint encoded I think it will take two bytes instead of just one (because the most significant bit of the first nibble `e` is 1). What do you think?

Kubuxu:
I am ok with it being 2bytes. It is just one byte extra and it is not used for content addressing that frequently (1 byte in case of sha would make a difference).

@whyrusleeping second opinion.

table.csv
```diff
@@ -200,3 +200,4 @@ stellar-tx,           Stellar Tx,                                       0xd1
 
 torrent-info,         Torrent file info field (bencoded),               0x7b
 torrent-file,         Torrent file (bencoded),                          0x7c
+ed25519,              Ed25519 public key,                               0xed
```
Kubuxu: I would change the name to `ed25519-pub`.

JustinDrake: Done 👍

lgierth:
I think this LGTM

PR #50, opened 2017-06-21
Add identity multihash

JustinDrake:
This is part of https://github.com/ipfs/go-ipfs/issues/3896

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -35,6 +35,7 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
+id,                 ,                         0x00
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
@@ -146,7 +147,6 @@ blake2s-232,,0xb25d
 blake2s-240,,0xb25e
 blake2s-248,,0xb25f
 blake2s-256,,0xb260
-id,                 ,                         0x50
 
 multiaddrs,,
 ip4,                ,                         0x04
--- a/table.csv
+++ b/table.csv
@@ -35,6 +35,7 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
+id,                 ,                         0x00
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
--- a/table.csv
+++ b/table.csv
@@ -35,7 +35,6 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
-id,                 ,                         0x00
 sha1,               ,                         0x11
 sha2-256,           ,                         0x12
 sha2-512,           ,                         0x13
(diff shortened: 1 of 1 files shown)


whyrusleeping:
cc @Kubuxu @diasdavid @lgierth 

We've talked about this for a while. `0x00` is the right number to use, yeah?

diasdavid:
@whyrusleeping could you point to that discussion. I remember the proposal but I'm not recalling exactly what were the main pros/cons

whyrusleeping:
@diasdavid https://github.com/multiformats/multihash/issues/13

lgierth:
I'm not sure we should use 0x00 codes. The general norm with string representation of bytes is that high zero bytes can be stripped. One prominent example of this is IPv6 adresses, where a `00ab` segment can and will be shortened to just `ab`. While it's technically fine because we can define the semantics of multicodecs ourselves, it goes against intuition and invites confusion and issues down the road.

We should treat the identity codec just as any other codec.

---

Previous discussions here: https://github.com/multiformats/multihash/issues/13 && https://github.com/multiformats/multicodec/pull/30

JustinDrake:
As a note to self, if we don't use the `0x00` code then [this merged PR](https://github.com/multiformats/go-multihash/pull/51/files) will have to be amended. It's a shame the `0x1d` code is taken, as a play on `id`.

diasdavid:
@lgierth brought up a good point. I see that @jbenet proposed 0x50 too? Would that work for everyone?

@JustinDrake yes, we would have to amend that part of the code.

JustinDrake:
I've amended this PR to use `0x50` instead of `0x00`.

lgierth:
So it turns out we're already using 0x00 for identity encoding in multibase :/

Let's stick with 0x00 here then, unless someone has a better idea. Sorry about that @JustinDrake :)

lgierth:
And let's also finally settle on ONE central codec table to rule them all (the one in the multicodec repo). These annoying inconsistency issues keep coming up and by now I really don't care about the arguments for one-table-in-every-repo.

lgierth:
Wait, before we carry on here: how is 0x00 gonna be represented in a string? Null-bytes in strings are a bad idea.

whyrusleeping:
@lgierth it will be converted to whatever encoding gets used. 

Though i don't know if the argument holds that we must use 0x00 here just because we use 0x00 for multibase. Then the binary multibase encoding of something that is using an 'identity' multicodec would start with two zero bytes.

diasdavid:
> Though I don't know if the argument holds that we must use 0x00 here just because we use 0x00 for multibase. 

Fair. And seems that here we always get saved most of the times by base encoding on top.

Let's go with 0x00 then

JustinDrake:
@lgierth No worries. I've reverted back to using `0x00` 👍

Stebalien:
@lgierth ping.

JustinDrake:
ping

lgierth:
Let's do it :+1: I'm still slightly concerned with starting with two zero-bytes in base2, but we don't avoid that by using something else here, it'll start with 0x00 one way or the other because of multibase.

lgierth:
> Though i don't know if the argument holds that we must use 0x00 here just because we use 0x00 for multibase.

Yeah it's not a very strong argument.

> Then the binary multibase encoding of something that is using an 'identity' multicodec would start with two zero bytes.

That's neat though!

Stebalien:
(long since fixed)

PR #52, opened 2017-07-20
Add p2p-circuit multiaddr

vyzo:


table.csv
```diff
@@ -162,6 +162,7 @@ http,               ,                         0x01E0
 https,              ,                         0x01BB
 ws,                 ,                         0x01DD
 onion,              ,                         0x01BC
+p2p-circuit,        ,                         0x0122
```
Stebalien: These should be sorted.

Stebalien: They aren't ... :cry:

vyzo: I can sort.

vyzo: Not sure we want to sort though :)

Stebalien: Yeah, just leave them as-is for now. We need to clean this all up in one go.

lgierth: They're sorted by topic :)

vyzo: so shall we merge?  I don't have perms on mulitformats.

PR #58, opened 2017-09-06
remove incorrect base2 code

kumavis:
not sure if there is a code for `base2` (ascii 101010101) yet but it is incorrectly set to the same as "raw binary"

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -5,7 +5,7 @@ bin,                raw binary,               0x55
 
 bases encodings,,
 base1,              unary,                    0x01
-base2,              binary (0 and 1),         0x55
+base2,              binary (0 and 1),         0x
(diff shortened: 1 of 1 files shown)


Stebalien:
The raw multicodec should be the null byte. This table looks correct.

kumavis:
@stebalien In the table (on master) "raw" and "base2" are currently 0x55. Which is correct?

Stebalien:
Ah, sorry, I got Raw (0x55) confused with IdentityMultihash/Base256/Binary (0x00). Base 2 is ASCII 0 (0x30).

Note: Raw is the raw dag format, why we have both is beyond me.

Also, all of these bases are messed up... We should probably record them as, e.g., `"0" (0x30)"`.

PR #59, opened 2017-09-07
[WIP] fill out multibase table

Stebalien:
DO NOT MERGE, we need to talk about this.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -1,20 +1,25 @@
 codec,              description,              code
 
 miscellaneous,,
-bin,                raw binary,               0x55
+bin,                raw binary,               0x55 # DUPLICATE CONFLICT
 
 bases encodings,,
-base1,              unary,                    0x01
-base2,              binary (0 and 1),         0x
-base8,              octal,                    0x07
-base10,             decimal,                  0x09
-base16,             hexadecimal,              0x
-base32,             rfc4648,                  0x
-base32hex,          rfc4648,                  0x
-base58flickr,       base58 flicker,           0x
-base58btc,          base58 bitcoin,           0x
-base64,             rfc4648,                  0x
-base64url,          rfc4648,                  0x
+identity,           raw binary,               0x00 ("\0") # DUPLICATE
+base1,              unary,                    0x31 ("1") # CONFLICT
+base2,              binary (0 and 1),         0x30 ("0") # CONFLICT
+base8,              octal,                    0x37 ("7")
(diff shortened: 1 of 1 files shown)


Stebalien:
@lgierth, @whyrusleeping, @kumavis, @jbenet 

So, we need to figure out how to make multibase a proper multicodec (or if we even want to do that). As-is, we have a lot of conflicts.

table.csv
```diff
@@ -1,20 +1,25 @@
...
-base64url,          rfc4648,                  0x
+identity,           raw binary,               0x00 ("\0") # DUPLICATE
+base1,              unary,                    0x31 ("1") # CONFLICT
+base2,              binary (0 and 1),         0x30 ("0") # CONFLICT
+base8,              octal,                    0x37 ("7")
+base10,             decimal,                  0x39 ("9")
+base16,             hexadecimal,              0x46 or 0x66 ("F" or "f")
+base32,             rfc4648,                  0x42 or 0x62 ("B" or "b")
+base32pad,          rfc4648,                  0x43 or 0x63 ("C" or "c")
+base32hex,          rfc4648,                  0x56 or 0x76 ("V" or "v") # CONFLICT
```
diasdavid: I believe a wrong assumption here is that the binary codec needs to be the value of the letter in ASCII. I mean, it would be nice, but there is nothing that tells us that.

Stebalien: That would mean we'd have to treat them as abstract symbols (although the binary one, "\0", is definitely `0x00`). We could do that but then I question whether or not they should appear on this list at all.

table.csv
```diff
@@ -1,20 +1,25 @@
...
-base8,              octal,                    0x07
-base10,             decimal,                  0x09
-base16,             hexadecimal,              0x
-base32,             rfc4648,                  0x
-base32hex,          rfc4648,                  0x
-base58flickr,       base58 flicker,           0x
-base58btc,          base58 bitcoin,           0x
-base64,             rfc4648,                  0x
-base64url,          rfc4648,                  0x
+identity,           raw binary,               0x00 ("\0") # DUPLICATE
```
Stebalien: We need to clarify the distinction between this and `bin` and if there even is one. If not, we may want to drop this one and make everything use `bin`. That would also affect the identity hash function still under discussion.

table.csv
Stebalien: Really, we just need to make some concrete decisions about this stuff.

PR #62, opened 2017-10-06
bin->raw

diasdavid:
When it got implemented in `go-ipfs`, devs called it raw https://github.com/ipfs/go-ipfs/blob/master/core/coredag/dagtransl.go#L22.

It should have followed the table (or change the table first). Now, to reduce friction and make both implementations interop, we will go and update js and the table.

@kumavis lead this efforted and created the PRs for it. Thanks @kumavis!

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -1,7 +1,7 @@
 codec,              description,              code
 
 miscellaneous,,
-bin,                raw binary,               0x55
+raw,                raw binary,               0x55
 
 bases encodings,,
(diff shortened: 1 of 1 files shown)


diasdavid:
Once this PR is approved, then:

- merge https://github.com/multiformats/js-multicodec/pull/19
- close https://github.com/ipld/js-ipld-raw/issues/2
- bubble up the changes to IPLD Resolver and IPFS.

PR #63, opened 2017-10-16
Aliases for protobuf and msgpack

crackcomm:


Stebalien:
What's the motivation here? This would prevent us from making a unique mapping from code to name.

crackcomm:
I rushed with this suggestion but after consideration it was never worth it. Not just because of the unique mapping but because of transparency, it's even more important. Thank you for your your feedback.

PR #65, opened 2017-10-24
add 0x00 for multibase binary

diasdavid:
//cc @whyrusleeping

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -5,7 +5,7 @@ raw,                raw binary,               0x55
 
 bases encodings,,
 base1,              unary,                    0x01
-base2,              binary (0 and 1),         0x
+base2,              binary (0 and 1),         0x00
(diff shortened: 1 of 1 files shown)


Kubuxu:
I think we were taking about not adding 0x00 as it would cause porblems with default values in Go. Also wasn't `base2` deemed a "joke"? As in it is number encoded as ASCI `10010` so ASCI string `100` would be 4.

Stebalien:
We really should distinguish between symbols and exact binary values in this table. The prefix for base2 is the symbol "0", not the byte 0x00 (`\0`). As @diasdavid and I discussed in #59, we need this distinction to disambiguate.

Note: 0x00 is *also* the identity multihash so `[0x00, data...]` is, in some respects, equivalent to `[0x55, data...]` (something that really annoys me...).

PR #68, opened 2017-12-29
fill out multibase table and treat multibases as symbols

Stebalien:
This extends the concept of multicodecs to general symbolic (text) strings, not just byte strings.

replaces #59

@diasdavid I've changed multibase prefixes to be "symbolic". Thoughts?

TODO: Deprecate 0x55 or 0x0? Having both is a bit annoying. They both mean "raw
binary data".

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -4,17 +4,27 @@ miscellaneous,,
 raw,                raw binary,               0x55
 
 bases encodings,,
-base1,              unary,                    0x01
-base2,              binary (0 and 1),         0x00
-base8,              octal,                    0x07
-base10,             decimal,                  0x09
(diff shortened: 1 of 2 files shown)


table.csv
```diff
@@ -35,6 +45,7 @@ multiaddr,          ,                         0x32
 multibase,          ,                         0x33
 
 multihashes,,
+identity,           raw binary,               0x0
```
diasdavid: Didn't we have a code for this one already?

Stebalien: We defined it [here](https://github.com/multiformats/multihash/blob/master/hashtable.csv) but never added that definition here.

Stebalien:
:steam_locomotive:

PR #72, opened 2018-02-01
Remove references to multicodec-packed

vmx:
multicodec-packed isn't a thing anymore, it was what multicodec is now.
Hence remove all references to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,7 @@
 
 - [Motivation](#motivation)
 - [How does it work? - Protocol Description](#how-does-it-work---protocol-description)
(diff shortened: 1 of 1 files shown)


Stebalien:
I believe we still have the concept of a "full" mutlicodec (IIRC, it was supposed to point to a file in IPFS) but maybe I'm misremembering. I believe we still use these in multistream.

However, given that we use multicodec to refer to the codec and never call it multicodec-packed, we *should* probably deprecate the term (and, if necessary, come up with a new one for the unpacked version).

@diasdavid you'd know more.

PR #73, opened 2018-03-19
Add multicodecs for Leofcoin

VandeurenGlenn:


table.csv
```diff
@@ -440,6 +440,10 @@ zcash-tx,             Zcash Tx,                                         0xc1
 stellar-block,        Stellar Block,                                    0xd0
 stellar-tx,           Stellar Tx,                                       0xd1
 
+leofcoin-block,       Leofcoin Block,                                   0x0A
+leofcoin-tx,          Leofcoin Transaction,                             0x54
+leofcoin-pr,          Leofcoin Peer Reputation,                         0x52
```
diasdavid: Cool! Is there an IPLD plugin too?

Kubuxu: I would prefer them being above `0x80` thus being in the two byte varint rage same as all other blockchains.

VandeurenGlenn: @diasdavid Almost, I discovered today that there is a interface-ipld-format repo so going to go through the readme there first before pushing.

VandeurenGlenn: Sure, do you have any preference?

Kubuxu: Not really, anything over 0x80 would be fine.

VandeurenGlenn:
Hi @diasdavid Maybe you know another way to test the plugin than changing dependency to my fork with the multicodecs added?

diasdavid:
@VandeurenGlenn note that this repo is just the spec, you still need to change it in go-multicodec or js-multicodec for it to take effect.

VandeurenGlenn:
@diasdavid That's the fork I was referring to, I should have been more clear, ok if it's only that what needs to happen then I probably overlooked something.

PR #79, opened 2018-07-30
table.csv: Add "zone" multiaddr (IPv6 zone)

mwnx:
Can be prefixed to the "ip6" protocol. See [rfc4007].

[rfc4007] https://tools.ietf.org/html/rfc4007


Stebalien:
Ok, we're doing this.

PR #82, opened 2018-09-03
Added common text formats

drew-512:
Added standard text formats


ntninja:
This has some overlap with #159 for actual text formats (rtf and plain are proposed here).

PR #85, opened 2018-10-06
add x11 hash algorithm

samli88:
This originated with Dash and is now used in multiple projects.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -63,6 +63,7 @@ keccak-256,         ,                         0x1B
 keccak-384,         ,                         0x1C
 keccak-512,         ,                         0x1D
 murmur3,            ,                         0x22
+x11,                ,                         0x23
 blake2b-8,Blake2b consists of 64 output lengths that give different hashes,0xb201
 blake2b-16,,0xb202
 blake2b-24,,0xb203


Stebalien:
Is this hash likely to be used in a context where shaving off a single byte is really important? If not, we may want to consider using a two-byte codec (>= 0x80). The other hash functions all use codecs that fit in a single byte as they're used in CIDs (content identifiers) where every byte really does count.

Stebalien:
> Can you help me understand how every byte counts for other hashes, e.g. murmur or dbl-sha2-256?

Well, for those two hash functions, it's probably not that important either. In retrospect, we probably should have chosen 2 byte codes for those as well. This really only applies to, e.g., sha256 (where hashes like this are used as identifiers).

I'm bringing this up because we only have 128 one byte codes available (due to the varint type we're using) and we're already using quite a few of them. I'd like to avoid using up even more of them unless size matters.

As for the byte sequence, it doesn't really matter as long as it's 2 bytes. We have 16256 total 2 byte codes so we're not pressed for space (at the moment).

samli88:
@Stebalien I pushed a new two-byte code for x11 hash. Will it work instead?

samli88:
I also just noticed https://github.com/multiformats/multihash/blob/master/hashtable.csv . Should it go there also? Seems to be two tables and not in sync.

Stebalien:
https://github.com/multiformats/multihash/pull/103

PR #92, opened 2018-11-16
Add Elixir implementation

nocursor:
First pass at an Elixir implementation of Multicodec.


vmx:
Thanks a lot for the Elixir implementation!

PR #93, opened 2018-11-26
reserve 0x2f (ASCII /) as the *path* multicodec

Stebalien:
That way, we can always tell if something is a path or something else.

We may also be able to take advantage of this later to combine a few
concepts and get rid of the "multiaddrs look like paths but are totally not"
problem. However, we can think about that later. This PR just reserves the code
so we don't run into problems later.

* Remove the distinction between string/binary multiaddrs. Instead, the "string"
  will *also* be a valid binary multiaddr.
* Define a new multipath spec to combine multiaddrs and other paths.

Related to: https://github.com/multiformats/multiaddr/issues/70


Stebalien:
@lgierth rebased.

PR #98, opened 2018-12-06
copy missing codecs from the multiaddr repo

Stebalien:
Also, rename ipfs to p2p as the *canonical* name.


Stebalien:
I've changed it to specify both libp2p and ipfs.

vmx:
We don't have an active maintainer at the moment, hence I dare to merge it.

PR #99, opened 2018-12-10
README: remove out-of-date captain

Stebalien:


vmx:
Although it would be cool to have a lead maintainer, it's better to have none than a non-active one.

PR #104, opened 2018-12-27
Add CIDs for IPFS and Swarm

Arachnid:
As discussed in https://github.com/multiformats/multiaddr/issues/73 and offline.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -447,3 +447,7 @@ leofcoin-pr,          Leofcoin Peer Reputation,                         0x83
 torrent-info,         Torrent file info field (bencoded),               0x7b
 torrent-file,         Torrent file (bencoded),                          0x7c
 ed25519-pub,          Ed25519 public key,                               0xed
+
+CIDs,,
+ipfs-cid,             IPFS CID,                                         0xee
+swarm-cid,            Swarm CID,                                        0xef
--- a/table.csv
+++ b/table.csv
@@ -449,5 +449,6 @@ torrent-file,         Torrent file (bencoded),                          0x7c
 ed25519-pub,          Ed25519 public key,                               0xed
 
 Content Namespaces,,
-ipfs-cid,             IPFS CID,                                         0xee
-swarm-cid,            Swarm CID,                                        0xef
+ipld-ns,              IPLD CID and optional path,                       0xee
+ipfs-ns,              IPFS CID and optional path,                       0xef
+swarm-ns,             Swarm CID and optional path,                      0xf0
--- a/table.csv
+++ b/table.csv
@@ -449,6 +449,6 @@ torrent-file,         Torrent file (bencoded),                          0x7c
 ed25519-pub,          Ed25519 public key,                               0xed
 
 Content Namespaces,,
-ipld-ns,              IPLD CID and optional path,                       0xee
-ipfs-ns,              IPFS CID and optional path,                       0xef
-swarm-ns,             Swarm CID and optional path,                      0xf0
+ipld-ns,              IPLD path,                                        0xee
+ipfs-ns,              IPFS path,                                        0xef
+swarm-ns,             Swarm path,                                       0xf0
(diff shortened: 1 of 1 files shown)


Arachnid:
Ping?

table.csv
```diff
@@ -447,3 +447,7 @@ leofcoin-pr,          Leofcoin Peer Reputation,                         0x83
 torrent-info,         Torrent file info field (bencoded),               0x7b
 torrent-file,         Torrent file (bencoded),                          0x7c
 ed25519-pub,          Ed25519 public key,                               0xed
+
+CIDs,,
```
Stebalien: Maybe "content namespaces"?. We're not really defining new types of CIDs.

1. ipld-ns: <ipld-ns-codec><cid>[/arbitrary/path]
2. ipfs-ns: <ipld-ns-codec><cid>[/arbitrary/path]
3. swarm-ns: <swarm-ns-codec><cid>[byte range?]

We could also merge #93 to allow arbitrary string paths (although we don't *have* to do it that way).

Basically, I'd like to make this as flexible as possible.

Arachnid: Can you elaborate on what an IPLD NS would encode, as opposed to an IPFS-NS? Would it be a CID, or some other data format?

Stebalien: I'm just taking string paths and encoding them in a compact binary form.

* `/ipld/SomeStringCID/some/remaining/path` -> `<ipld-namespace-multicodec><binary-cid><path>`
* `/ipfs/SomeStringCID/some/remaining/path` -> `<ipfs-namespace-multicodec><binary-cid><path>`

So these will both be CIDs + Paths.

We need different namespaces because IPFS is built on top of IPLD. In IPFS, we' interpret the content as files and directories and the path a traversal through a directory tree. In IPLD we interpret the content as structured data and the path as a traversal through a JSON-like datastructure.

---

For the swarm namespace, I wrote `[byte range?]` as you *may* want to allow for a byte range in addition to the CID.

Stebalien: As a string although we may want to length-prefix these strings.

Regardless, we can probably skip that for now. The important part here is to reserve these numbers. Regardless of what we say now, we'll probably change our minds when we actually try to prototype these.

table.csv
```diff
@@ -447,3 +447,8 @@ leofcoin-pr,          Leofcoin Peer Reputation,                         0x83
 torrent-info,         Torrent file info field (bencoded),               0x7b
 torrent-file,         Torrent file (bencoded),                          0x7c
 ed25519-pub,          Ed25519 public key,                               0xed
+
+Content Namespaces,,
+ipld-ns,              IPLD CID and optional path,                       0xee
```
Stebalien: We can be a bit more abstract here and just say "IPLD path", "IPFS path", and "Swarm path" instead of specifying the precise format for the argument. That is, we know we'll use `<namespace-codec><stuff...>` but we can decide on `<stuff...>` later.

For example, swarm probably won't want arbitrary paths (it doesn't support directories, as far as I know. Instead, it will at *most* want byte ranges (e.g., `/swarm/CID` or `/swarm/CID/0..200`).

Arachnid: Swarm does support paths, in a similar fashion to IPFS, by storing manifest files.

Stebalien: Ah, got it. I just saw the docs on Swarm-Hash and assumed it was blob-only.

Stebalien: In that case, I there is some value in trying to standardize on a binary format for content paths up-front. My thinking was just `<namespace-codec (varint)><binary-cid><string-path>` (where `<string-path>` is an optional `/a/b/c...`). Thoughts?

(we can also discuss this in a separate issue if you'd like).

Arachnid: That seems reasonable. I liked the suggestion of reserving `/` as the multicodec for "path component". I think this'd be best discussed in its  own issue, though.

Stebalien:
I had one job... I guess it's time to give in to our robot overlords.

Fix and CI in: https://github.com/multiformats/multicodec/pull/113

PR #114, opened 2019-01-24
Use '#!/usr/bin/env python3' vs. fixed path '/usr/bin/python3'

pjkundert:
Many installations have python3 installed elsewhere, such as /usr/local/bin/, etc.  Use the environment PATH variable to locate, using /usr/bin/env.


Stebalien:
Thanks!

PR #117, opened 2019-01-30
Update table.csv

keutmann:
Added the RIPEMD cryptographic hash functions 
RIPEMD-160 is the base hash encoding for Bitcoin addresses.


Stebalien:
So, I'm a bit worried about allocating 4 more 1 byte codes (<= 0x80). We have a limited number (and should have been more careful in the past).

What if we allocated ripemd-160 in this range and everything else in the two byte range?

keutmann:
Excellent suggestion, so fine by me :)

PR #120, opened 2019-02-20
Add Swarm multicodecs

justelad:
Hi,

I'd like to add these two Swarm primitive types that we'd like to be able to reference from ENS resolver contracts.

Please let me know if I should shift the codes around (although obviously lower numbers = less data on chain).

Cheers

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -68,6 +68,8 @@ swarm-ns,                       namespace,      0xe4,           Swarm path
 ed25519-pub,                    key,            0xed,           Ed25519 public key
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
+swarm-manifest,                 swarm,          0xfa,           Swarm Manifest
+swarm-feed,                     swarm,          0xfb,           Swarm Feed
 udp,                            multiaddr,      0x0111,
 p2p-webrtc-star,                multiaddr,      0x0113,
 p2p-webrtc-direct,              multiaddr,      0x0114,
--- a/table.csv
+++ b/table.csv
@@ -60,6 +60,7 @@ stellar-block,                  ipld,           0xd0,           Stellar Block
 stellar-tx,                     ipld,           0xd1,           Stellar Tx
 md4,                            multihash,      0xd4,
 md5,                            multihash,      0xd5,
+bmt,                            multihash,      0xd6,           Binary Merkle Tree Hash
 decred-block,                   ipld,           0xe0,           Decred Block
 decred-tx,                      ipld,           0xe1,           Decred Tx
 ipld-ns,                        namespace,      0xe2,           IPLD path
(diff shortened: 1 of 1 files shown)


Stebalien:
So we can correctly categorize these, how are they being used?

justelad:
@Stebalien These are Swarm notions of base layer primitives that resolve automagically to actual content.

My assumption is that users would want to point to either one of them when setting an ENS resolver contract content hash as discussed in EIP-1577.

So if I understand the question, they will be used by users to point to content. 
In general I don't see any need to represent any other Swarm notions inside multicodecs as we define file metadata inside manifests; not as a self defining hash. So in fact the purpose of this PR is to define the standard of how to set content hashes in the revised ENS resolvers.

justelad:
Also, if I may ask to put us in an address space where we could have a few more subsequent future entries reserved.

Many thanks

justelad:
Should I revise anything in this PR?

Stebalien:
So, are these used to identify the type of content? I'm asking because I _think_ these should be categorized as IPLD codecs: they tell you how to understand the addressed content (which I'm guessing isn't self-describing).

Stebalien:
On the original swarm PR, there was a lot of discussion around using CIDs. That is, ENS records would have values of the format `<namespace (multicodec)><CID><path?>`. For example, `<swarm-ns><CID (codec=swarm-manifest)>/path/to/file`.

justelad:
@Stebalien I suspect I'm not 100% following you and I'm not 100% sure how CID/Multicodecs/IPLD come into play altogether in this case.

I guess that the decision whether to classify these as IPLD codecs is up to you (aren't IPLD codecs using multicodecs anyway? why does it even matter if we specify them as one codec or another? I'm afraid I don't really understand the relationships between the different (many) moving parts in the multiformat/ipld specs). The discussions that were made on the topic were not very clear from our side, not to mention that a significant part of them was conducted off-band without any documentation or resolution. 

Please advise, and in any case it would be nice to hop on a call and maybe say hello and work this out more quickly as we want to merge the new ENS public resolver changes.

justelad:
@Stebalien, I had a talk with @zelig and we have concluded that we'd like to add a new `multihash` code for `bmt` (as Swarm hashes are actually binary merkle tree hashes, not `keccak256` hashes per-se).

You could [see](https://github.com/ethersphere/go-ethereum/pull/389/files) how we've added local support for that at the time in Swarm (as of a few months ago we have dropped usage of `multiformats`, but this will be reintroduced with the changes to the resolver).

Stebalien:
Sorry, github notifications tend to fall through the cracks. This looks good to me.

With respect to IPLD: IPLD is really just a (poorly documented...) attempt at unifying all merkle formats into a single system. It does so by using CIDs (content identifiers) which are `(multicodec, multihash)` tuples (where the multicodec == "ipld codec"). This allows us to point to an object in a merkle dag while recording the _type_ of merkledag we're pointing to.

TL;DR: IPLD is to merkledags as Multihash is to hash functions.

PR #121, opened 2019-03-17
Add multicipher & AES-256-CBC to multicodec table

tboeckmann:
* multicipher for cipher/ encrypted formats
* AES-256-CBC (a multicipher) for file encryption


tboeckmann:
This proposes `multicipher` for encryption formats.
We're using AES-256-CBC encryption on IPFS stored files and looking to standardise the multiformat table entry.
Note: the file location is sometimes added to TX data, so the shortest possible code is required to reduce TX data length.

@Stebalien - any thoughts here?

lgierth:
Where can I read more about multicipher? I found multiple projects called multicipher but none of them look related.

tboeckmann:
@lgierth Multicipher refers to the use of multiple ciphers in a program. For example, in https://github.com/mailchain/mailchain#multicipher, default encryption is AES-256-CBC, but other ciphers can be used.

PR #122, opened 2019-03-21
Reserve multicodec for filecoin compact serialization format

whyrusleeping:
Ref: https://github.com/filecoin-project/specs/pull/131

Note that we're only using one multicodec here, not one per object type.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -15,6 +15,7 @@ keccak-224,                     multihash,      0x1A,           keccak has varia
 keccak-256,                     multihash,      0x1B,
 keccak-384,                     multihash,      0x1C,
 keccak-512,                     multihash,      0x1D,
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -15,6 +15,7 @@ keccak-224,                     multihash,      0x1A,           keccak has varia
 keccak-256,                     multihash,      0x1B,
 keccak-384,                     multihash,      0x1C,
 keccak-512,                     multihash,      0x1D,
+filecoin compact serialization, ipld,           0x1F,
```
Stebalien: Name should be something like `filecoin` (or `filecoin-object`?). The description can go in the fourth field.

rvagg:
https://github.com/multiformats/multicodec/pull/161#issuecomment-627152617

PR #123, opened 2019-03-26
Add MessagePack serialization format.

Kenadia:


Stebalien:
Sorry for the rather late review. Does this _need_ a one-byte code or can we use a 2 byte code (0x80+)? We have at most 127 of them and we've already wasted too many without thinking.

Kenadia:
I thought MessagePack is popular enough of a format that it was best placed with all of the other serialization formats.

I'm fine putting it anywhere though. I'll try 0x0200 unless you have another preference.

Stebalien:
> I thought MessagePack is popular enough of a format that it was best placed with all of the other serialization formats.

Unfortunately, it's not about _placement_, it's just that we have 127 codes to assign and have already assigned way too many. I'd prefer to move out 90% of the codes in that range but it may be a bit late for that.

0x0200 works.

Stebalien:
Basically, we try to reserve single-byte codes for codes used in addresses (multiaddr, ipld, and multihash). Really, cbor, protobuf, rlp, and bencode should be moved elsewhere.

Stebalien:
(and can you rebase on master?)

Kenadia:
Got it, thanks for the explanation!

I noticed JSON is also missing and I was wondering how you feel about adding that first:
https://github.com/multiformats/multicodec/pull/141

Stebalien:
Merge the json one.

Kenadia:
Thanks, I've updated this PR to use 0x0201

PR #124, opened 2019-03-27
remove bmt in favor of using keccak256

justelad:
Hi there,

In light of discussions with ENS maintainers we have decided to abandon the use of `bmt` in favor of using plain `keccak256` multihash codes to identify swarm hashes.

Thanks!

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -60,7 +60,6 @@ stellar-block,                  ipld,           0xd0,           Stellar Block
 stellar-tx,                     ipld,           0xd1,           Stellar Tx
 md4,                            multihash,      0xd4,
 md5,                            multihash,      0xd5,
-bmt,                            multihash,      0xd6,           Binary Merkle Tree Hash
 decred-block,                   ipld,           0xe0,           Decred Block
 decred-tx,                      ipld,           0xe1,           Decred Tx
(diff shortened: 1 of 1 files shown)


vmx:
@Stebalien Do we have a policy on removing things? Would we just keep everything so that they can't be re-claimed?

justelad:
We can also keep it for the time being. But then again the libraries that implement multicodecs have not yet been updated with bmt.

lgierth:
Is it still used in the Eth ecosystem or elsewhere? If so, let's keep it.

If it's dead now that ENS decided for Keccak, let's drop it.

vmx:
BMT was just added recently (less than a month ago), so I'd be OK with just removing it again.

vmx:
@justelad it would be great if you could comment on @lgierth's questions: https://github.com/multiformats/multicodec/pull/124#issuecomment-477176055

As I expect it to be a "it's dead now", @Stebalien would you be OK with removing it if that's the case?

justelad:
hey.
apologies for stagnating on this. I've had a chance to sync up with other people in the foundation and apparently there's a hanging issue to integrate and implement `bmt` in solidity. so let's reserve `bmt` for future use.
thanks! 

@vmx regarding the ENS issue we're still going with keccak at the moment

PR #125, opened 2019-03-29
Add ECDSA

mathiasrw:
Elliptic Curve Digital Signature Algorithm or ECDSA is a cryptographic algorithm used by Bitcoin to ensure that funds can only be spent by their rightful owners.


table.csv
```diff
@@ -67,6 +67,7 @@ decred-tx,                      ipld,           0xe1,           Decred Tx
 ipld-ns,                        namespace,      0xe2,           IPLD path
 ipfs-ns,                        namespace,      0xe3,           IPFS path
 swarm-ns,                       namespace,      0xe4,           Swarm path
+ECDSA,                          multihash,      0xec,           ECDSA signature
```
Stebalien: This wouldn't be a multihash, it would be a signature. So we're on the same page, how are you planning on using this codec to identify ECDSA signatures?

mikeal: do we have plans on doing a “multisign” at some point?

Stebalien: We do but it's unclear how this should look. The signature algorithm is usually a property of the _key_, not the signature, and libp2p actually has an ECDSA key type.

blabno: I'm working on clojure implementation of multisig and multikey.
Multikey is for expressing public keys and would look like this: algo+length+x509encodedKey
Where `algo` is ecdsa, rsa, or other, `length` is length of `x509encodedKey`.

Multisig is for expressing signatures and would look like this: algo+length+signatureBytes
Where `algo` could be `sha256-with-ecdsa`.

Unlike with multihashes, it would not be possible to compare payloads for equality (as signature may be different for the same data each time the signature is calculated), but having a public key it would still be possible to verify the signature.

awoie: IMO, ECDSA is a bit too generic. For example, how do you distinguish between secp256k1 and r1?

blabno: Do you still need curve params if you have x509encodedKey as part of multikey?

PR #126, opened 2019-04-05
Add garlic32, describe garlic addresses

eyedeekay:
This goes with the other I2P base32 pull requests, it adds i2p base32 addresses and adds descriptions of what the I2P Addresses are to the table.


Stebalien:
Can we call this "i2p base64 (raw public key)" and "i2p base32 (hashed public key)"? Every time I see "base64" versus "base32" I get confused and re-ask myself why we aren't just using multibase. Then I remember that we're just using the base encodings to distinguish between two very different address formats.

eyedeekay:
No problem. Technically it's just *slightly* inaccurate in some cases because that's not all the new base32's (for encrypted leasesets) have, but that is the default and far and away the most common case.

PR #128, opened 2019-04-21
Normalize hexadecimal syntax of codes

nichtich:
This commit normalizes hexadecimal syntax of codes to lowercase and ensures that uppercase and lowercase will not be mixed again.


lgierth:
Looks good, thanks :)

PR #129, opened 2019-04-22
Add git blob object multihash

nichtich:
The git version control system identifies file contents with object
hashes. The hash is calculated by prepending the character sequence
"blob ", the length in bytes as human readable integer, and a NUL byte
before calculating the SHA-1 hash. There are plans to migrate from SHA1
to SHA256, so the hash name `git-sha1` should indicate the hash
function. In addition to "binary large object" hashes git has hashes for
"commit", "tree", and "tag" object types but these are only of interest
in context of a git repository while git blob object hashes can be used
on arbitrary digital objects.

Example:

    $ echo -n "Merkle–Damgård" | git hash-object -w --stdin
    8c85276d1af7f277e82bbf98ddaf30ff3644bcb1

    $ echo -n "blob 17\0Merkle–Damgård" | shasum
    8c85276d1af7f277e82bbf98ddaf30ff3644bcb1  -

To retrieve objects from a git repository by their hash, use `git cat-file`

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -62,6 +62,7 @@ stellar-tx,                     ipld,           0xd1,           Stellar Tx
 md4,                            multihash,      0xd4,
 md5,                            multihash,      0xd5,
 bmt,                            multihash,      0xd6,           Binary Merkle Tree Hash
+git-sha1,                       multihash,      0xd7,           git blob object hash (based on SHA1)
 decred-block,                   ipld,           0xe0,           Decred Block
 decred-tx,                      ipld,           0xe1,           Decred Tx
 ipld-ns,                        namespace,      0xe2,           IPLD path


magik6k:
There is already the `git-raw` codec, which can be used for blobs too. The hash function is specified in the multihash header part of CIDs.

There also exists an implementation of this codec - https://github.com/ipfs/go-ipld-git/blob/master/git.go#L47

nichtich:
> There is already the `git-raw` codec, which can be used for blobs too.

Code `0x78` with name `git-raw` is not defined as `multihash` but as `ipld`, so it's not part of the [multihash registry](https://w3c-dvcg.github.io/multihash/). As far as I understand the IPLD implementation in go, the use case is different. The current table also contains two codes for "raw binary", `0x00` (multihash) and `0x55` (ipld), so both seem to be different. I must admit I'm not familiar with IPLD. My use case is generation of a hash from an arbitrary sequence of bytes independently from IPLD and/or IPFS.

Maybe @msporny can help out to clarify whether and how ipld can be used for multihash? Following my addition of code `0x7d`, the multihash of the string "Merkle–Damgård" would be 0xd7208c85276d1af7f277e82bbf98ddaf30ff3644bcb1. With code `0x78` there is no valid multihash but maybe the [multihash specification](https://github.com/w3c-dvcg/multihash) can be modified to also include `0x78` for git blob sha1 hashes, so the resulting multihash of the string "Merkle–Damgård" would be 0x78208c85276d1af7f277e82bbf98ddaf30ff3644bcb1.

magik6k:
What you'd want to use in this situation is https://github.com/multiformats/cid, which combines multicodec+multihash.

Multihash is just concerned with hash algorithms, which in case of git is just a sha1 sum of raw (uncompressed) objects (sha1 code is ` 0x11`). Multicodec is what we use to relay content-type information (git object, ipfs protobuf block, etc.), but it's not tied to hash algorithms. CIDs merge those concepts and allow you to say `git-raw - sha1`.

We could consider adding separate codecs for all the different object types as afaik there wasn't really any discussion on what is the correct thing to do.

nichtich:
Thanks for the pointer to CID, but CID adds content type (multicodec). git instead has no knowledge about content types, a git blob hash can index any kind of byte string (=any file content). This way a git repository is also a content-adressable file system. The git blob hash is internally based on SHA1 but its hash value obviously differs from SHA1.

magik6k:
The multicodec specified in this context doesn't map to anything in git, it only says 'this is a git object'. If you want to create git CIDs/multihashes, the git-raw codec should take care of prepending the blob header for you, hash function shouldn't be involved in that. 

If you could share a bit more about your usecase it might make it a bit easier to reason about.

Stebalien:
To be explicit, the CID codec is explicitly _not_ a "content" type. It's a _merkletree_ type.

Stebalien:
(closing for tracking, feel free to continue discussing)

PR #134, opened 2019-05-29
Make README reflect what multicodec currently is used for

### LINKED ISSUE #133: Outdated README
The current README is heavily outdated. You can see that there's several left-overs from earlier versions of the multiformats stuff (e.g. references to multistreams). The current README defines multicodec as being a data format/encoding like the other multiformats are. It's specified as `<codec-encoded-as-varint><data>`.

I went over the other multiformats (multiaddr, multihash, multibase, multistream-select, cid). There the term "multicodec" is used (if used) as being am identifier encoded as varint, hence just as `<codec-encoded-as-varint>` without the `<data>` piece.

I can see two ways moving forward. We could expand the multicodec format and use it also in other specs, or we drop the multicodec format and use just the codec table.

### Expanding the multicodec format

Currently the only case where we use `<codec-encoded-as-varint><data>` is within CIDs. To expand it to other specs it would need to change a bit to be useful. It could be `<codec-encoded-as-varint><data-length-as-varint><data>` with optionally just `<codec-encoded-as-varint><data>` if the codec specifies a fixed size (e.g. multiaddr when using ipv4).

That way multihash would just be a multicodec. Also the binary representation of multiaddr would just be a list of multicodecs.

### Dropping multiformat as format

We could also just remove the `<codec-encoded-as-varint><data>` part and define multicodec as just being a table of codecs which is used by the various multiformats.

If that's the plan I'd suggest renaming it to "multiformat codecs" and moving the table into the multiformats repo. Then removing the multicodec repository as it would be meaningless.

/cc @mikeal

vmx:
Multicodec changed over time what it actually is. The README should reflect
the current state on how it is used.

Closes #133.

Please see this as a starting point we can hopefully find agreement on. Please
also help to rephrase my changes into better English (I couldn't get rid of overusing
the word "use").

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,54 +5,35 @@
 [![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](https://webchat.freenode.net/?channels=%23ipfs)
 [![](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/RichardLitt/standard-readme)
 
-> Compact self-describing codecs. Save space by using predefined multicodec tables.
+> Canonical table of of codecs used by various multiformats
 
 ## Table of Contents
 
 - [Motivation](#motivation)
-- [How does it work? - Protocol Description](#how-does-it-work---protocol-description)
+- [Description](#description)
+- [Examples](#examples)
 - [Multicodec table](#multicodec-table)
+  - [Adding new multicodecs to the table](#adding-new-multicodecs-to-the-table)
 - [Implementations](#implementations)
 - [FAQ](#faq)
-- [Maintainers](#maintainers)
 - [Contribute](#contribute)
 - [License](#license)
 
 ## Motivation
 
-[Multistreams](https://github.com/multiformats/multistream) are self-describing protocol/encoding streams. Multicodec uses an agreed-upon "protocol table". It is designed for use in short strings, such as keys or identifiers (i.e [CID](https://github.com/ipld/cid)).
+Multicodec is an agreed-upon codec table. It is designed for use in binary representations, such as keys or identifiers (i.e [CID](https://github.com/ipld/cid)).
 
-## Protocol Description - How does the protocol work?
+## Description
 
-`multicodec` is a _self-describing multiformat_, it wraps other formats with a tiny bit of self-description. A multicodec identifier is a varint.
+The code of a multicodec is usually encoded as unsigned varint as defined by [multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint). It is then used as a prefix to identify the data that follows.
 
-A chunk of data identified by multicodec will look like this:
+## Examples
 
-```sh
-<multicodec><encoded-data>
-# To reduce the cognitive load, we sometimes might write the same line as:
-<mc><data>
-```
-
-Another useful scenario is when using the multicodec as part of the keys to access data, example:
-
-```
-# suppose we have a value and a key to retrieve it
-"<key>" -> <value>
-
-# we can use multicodec with the key to know what codec the value is in
-"<mc><key>" -> <value>
-```
-
-It is worth noting that multicodec works very well in conjunction with [multihash](https://github.com/multiformats/multihash) and [multiaddr](https://github.com/multiformats/multiaddr), as you can prefix those values with a multicodec to tell what they are.
-
-## MulticodecProtocol Tables
-
-Multicodec uses "protocol tables" to agree upon the mapping from one multicodec code. These tables can be application specific, though -- like [with](https://github.com/multiformats/multihash) other [multiformats](https://github.com/multiformats/multiaddr) -- we will keep a globally agreed upon table with common protocols and formats.
+Multicodec is used in various [Multiformats](https://github.com/multiformats/multiformats). In [Multihash](https://github.com/multiformats/multihash) it is used to identify the hashes, in the machine-readable [Multiaddr](https://github.com/multiformats/multiaddr) to identify components such as IP addresses, domain names, identities, etc.
 
 ## Multicodec table
 
-The full table can be found at [table.csv](/table.csv) inside this repo. There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
+Find the canonical table of multicodecs at [table.csv](/table.csv). There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
 
 ### Adding new multicodecs to the table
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,54 +5,35 @@
 [![](https://img.shields.io/badge/freenode-%23ipfs-blue.svg?style=flat-square)](https://webchat.freenode.net/?channels=%23ipfs)
 [![](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg?style=flat-square)](https://github.com/RichardLitt/standard-readme)
 
-> Compact self-describing codecs. Save space by using predefined multicodec tables.
+> Save space by using the predefined multicodec table
```
raulk: This is a *claim*. I think the subtitle should reflect what multicodec *is*. I'd describe it as a canonical table enumerating well-known content types.

README.md
```diff
@@ -5,54 +5,35 @@
...
 - [Implementations](#implementations)
 - [FAQ](#faq)
-- [Maintainers](#maintainers)
 - [Contribute](#contribute)
 - [License](#license)
 
 ## Motivation
 
-[Multistreams](https://github.com/multiformats/multistream) are self-describing protocol/encoding streams. Multicodec uses an agreed-upon "protocol table". It is designed for use in short strings, such as keys or identifiers (i.e [CID](https://github.com/ipld/cid)).
+Multicodec is an agreed-upon codec table. It is designed for use in binary representations, such as keys or identifiers (i.e [CID](https://github.com/ipld/cid)).
```
raulk: What's the definition of a "codec"? We're using it in an unconventional way. Usually the codec is a device or software that performs encoding and decoding.

README.md
```diff
@@ -5,54 +5,35 @@
...
 
-## Protocol Description - How does the protocol work?
+## Description
 
-`multicodec` is a _self-describing multiformat_, it wraps other formats with a tiny bit of self-description. A multicodec identifier is a varint.
+The code of a multicodec is usually encoded as unsigned varint as defined by [multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint). It is then used as a prefix to identify the data that follows.
```
raulk: Let's remove usages from the description. Prefixing data with the multicodec is a valid use case, but not the only one. We can add a section on "Known/prescribed usages".

README.md
```diff
@@ -5,54 +5,35 @@
...
-# we can use multicodec with the key to know what codec the value is in
-"<mc><key>" -> <value>
-```
-
-It is worth noting that multicodec works very well in conjunction with [multihash](https://github.com/multiformats/multihash) and [multiaddr](https://github.com/multiformats/multiaddr), as you can prefix those values with a multicodec to tell what they are.
-
-## MulticodecProtocol Tables
-
-Multicodec uses "protocol tables" to agree upon the mapping from one multicodec code. These tables can be application specific, though -- like [with](https://github.com/multiformats/multihash) other [multiformats](https://github.com/multiformats/multiaddr) -- we will keep a globally agreed upon table with common protocols and formats.
+Multicodec is used in various [Multiformats](https://github.com/multiformats/multiformats). In [Multihash](https://github.com/multiformats/multihash) it is used to identify the hashes, in the machine-readable [Multiaddr](https://github.com/multiformats/multiaddr) to identify the protocol.
```
raulk:
```suggestion
Multicodec is used in various [Multiformats](https://github.com/multiformats/multiformats). In [Multihash](https://github.com/multiformats/multihash) it is used to identify the hashes, in the machine-readable [Multiaddr](https://github.com/multiformats/multiaddr) to identify components such as IP addresses, domain names, identities, etc.
```

README.md
```diff
@@ -5,54 +5,35 @@
...
 
 ## Multicodec table
 
-The full table can be found at [table.csv](/table.csv) inside this repo. There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
+As multicodecs are used across several multiformats, there's now a single table for all of them which can be fount at [table.csv](/table.csv) inside this repo. There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
```
raulk: We should drop language that refers to an elder time, as the reader won't be interested in the past but rather the present.

```suggestion
Find the canonical table of multicodecs at [table.csv](/table.csv). There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
```

README.md
mikeal: In the case of IPLD that’s exactly what codecs do, they encode and decode blocks.

README.md
vmx: What about something like "Canonical table of of codecs used by various multiformats". I still don't like that one, if anyone can come up with something better, that would be great.

README.md
vmx: @Stebalien Do I recall correctly that you had a good answer to the question, why it is "codec" and not "content type". If yes, would you mind adding that as a FAQ?

README.md
vmx: @raulk: Do you mean removing the whole thing, or just the last sentence.

vmx: @raulk What are other use cases?

vmx:
@Stebalien I haven't seen any comments from you yet, it would be great to hear yours (as you are one of the defacto maintainers of multicodec).

vmx:
I think I've addressed all comments.

Can we find agreement that with merging this PR things are better than before (even if things are not perfect). If yes then please approve this PR so it can get merged.

SleeplessByte:
This completely removes the [protocol description](https://github.com/multiformats/multicodec#protocol-description---how-does-the-protocol-work). Reading through the other changes, isn't that section still valid?

In `multihash` it is used exactly as the first example: the multicodec key describes how the hash was created.
In `cid` it used like the "key" -> "value" example: the multicodec key describes the *content* that is addressed (which would be the value if it was a cid -> content mapping).

vmx:
@SleeplessByte Yes that's intentional. The point is that multicodec is now just the codecs, not a protocol anymore. Multihash is mentioned as an example on how you can use multicodec in a protocol like Multihash.

SleeplessByte:
Got it, thank you for the elaboration :heart:. I think it would be _really_ helpful to add that **exact** language.

vmx:
I've rebased the changes to current master (and fixed up the commits).

@Stebalien I think this version is better than what we currently have, what do you think?

vmx:
@mikeal @raulk @Stebalien: This PR hasn't see much activity recently. Could you please have another look and respond to my questions or re-raise your concerns? I still find the version of this PR better than what we currently have.

README.md
mikeal: content type has an existing meaning, and if you use that meaning then you could apply multiple content types to many codecs, and multiple codecs to a single content type, which means they cannot be synonymous.

Example: A block of `dag-json` can also be interpreted using `json` and the semantic meaning of the links would b stripped. The “content type” is still JSON. Similarly, you could interpret as `raw` and lose all semantic meaning. Content type is essentially what we’ve tried to describe as a “format” but unfortunately there’s a lot of older libraries and documentation that use the term format interchangeably with codec.

To add the confusion if we were to re-use content type, we already need to apply content types in UnixFS but we have to apply them to a “file” which is actually a dag connecting many blocks of raw binary data. Nowhere in that graph would you be able to use a content-type in place of the codec as a correct signal, it has to be added to the File metadata, or potentially wrapped in content-type metadata (the UnixFSv1 approach).

In short, a codec tells you how to decode block data into IPLD Data Model (JSON types + Links) and a content type tells an application how to interpret data (which may or may not involve parsing). There’s certainly some overlap, video file content types signal a literal codec for interpretation, but our interpretation is limited to how to interpret a single block into Data Model.

mikeal:
LGTM.

vmx:
It's been approved, there are no recent objections, hence merging it.

PR #136, opened 2019-06-09
Add IPNS namespace

mcdee:
Add an IPNS namespace.  This will have the same format as IPFS but allow resolution of IPNS paths.


Amxx:
#137

mcdee:
@Stebalien any chance of merging this?  Making the IPNS namespace official will help with the ENS integration.

PR #138, opened 2019-06-20
Add Ruby implementation

SleeplessByte:
I've implemented this as proposed in #134, which means that the ruby implementation is just that: a table with some convenience. Currently it would be up to the ruby versions of cid (wip) and outdated ruby multihash to use these tables and add/intepret the prefixes.


vmx:
@SleeplessByte Sorry I somehow missed that PR. Thanks again for working on the Ruby stuff.

PR #140, opened 2019-07-23
Add ZeroNet site address

filips123:
Add [ZeroNet](https://zeronet.io/) site address to protocol codes. ZeroNet uses old-style Bitcoin addresses as site addresses.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -74,7 +74,6 @@ dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
 swarm-manifest,                 ipld,           0xfa,           Swarm Manifest
 swarm-feed,                     ipld,           0xfb,           Swarm Feed
-zeronet,                        multiaddr,      0x1000,         ZeroNet site address (old-style Bitcoin address)
 udp,                            multiaddr,      0x0111,
 p2p-webrtc-star,                multiaddr,      0x0113,
 p2p-webrtc-direct,              multiaddr,      0x0114,
@@ -96,6 +95,7 @@ ws,                             multiaddr,      0x01dd,
 wss,                            multiaddr,      0x01de,
 p2p-websocket-star,             multiaddr,      0x01df,
 http,                           multiaddr,      0x01e0,
+zeronet,                        namespace,      0x1000,         ZeroNet site address
 x11,                            multihash,      0x1100,
 blake2b-8,                      multihash,      0xb201,         Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,
--- a/table.csv
+++ b/table.csv
@@ -69,6 +69,7 @@ ipld-ns,                        namespace,      0xe2,           IPLD path
 ipfs-ns,                        namespace,      0xe3,           IPFS path
 swarm-ns,                       namespace,      0xe4,           Swarm path
 ipns-ns,                        namespace,      0xe5,           IPNS path
+zeronet,                        namespace,      0xe6,           ZeroNet site address
 ed25519-pub,                    key,            0xed,           Ed25519 public key
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
@@ -97,7 +98,6 @@ p2p-websocket-star,             multiaddr,      0x01df,
 http,                           multiaddr,      0x01e0,
 json,                           serialization,  0x0200,         JSON (UTF-8-encoded)
 messagepack,                    serialization,  0x0201,         MessagePack
-zeronet,                        namespace,      0x1000,         ZeroNet site address
 x11,                            multihash,      0x1100,
 blake2b-8,                      multihash,      0xb201,         Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,
--- a/table.csv
+++ b/table.csv
@@ -74,6 +74,7 @@ dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
 swarm-manifest,                 ipld,           0xfa,           Swarm Manifest
 swarm-feed,                     ipld,           0xfb,           Swarm Feed
+zeronet,                        multiaddr,      0x1000,         ZeroNet site address (old-style Bitcoin address)
 udp,                            multiaddr,      0x0111,
 p2p-webrtc-star,                multiaddr,      0x0113,
 p2p-webrtc-direct,              multiaddr,      0x0114,


Stebalien:
Is a ZeroNet address used to address a process/machine? A multiaddr tells you how to talk to a peer. I believe we may want an IPLD codec/CID.

How are you planning on using this codec (i.e., by example).

filips123:
The ZeroNet address is used to represent a website hosted on the ZeroNet network. This works in a similar way as you use IPFS/IPNS hash to represent file hosted on IPFS network (except that ZeroNet address is same even after file changes).

I would need this codec to implement ENS domain support in ZeroNet. I need to get `contenthash` record type from ENS domain. That records are using multicodec and multiaddr to differentiate between multiple types and encode them correctly. This is why I need ZeroNet address codec.

Stebalien:
As @alexander225 noted in https://github.com/multiformats/multiaddr/pull/97#issuecomment-515249005, this isn't really a "multiaddr".

We have two types of codecs that may be appropriate:

* Namespace: Given something like, e.g., /swarm/CID/... or /ipfs/CID/..., the _namespace_ codec would encode the `/swarm/` and `/ipld/` parts, telling us us which system the data likely lives in and how to path through it.
* IPLD: With an IPLD codec, you'd be able to create a [CID](https://github.com/multiformats/cid) that indicates how to understand the referenced data as structured, merkle-linked data.

How does resolution work precisely?

* What is the format of the key
* What is the format of the signature(?) on the content.
* What is the format/encoding of the referenced content? What is the format of the data being signed?

For example, in IPNS,

* The key is an libp2p public key.
* The signature format depends on the public key type.
* The data being signed is an arbitrary string path (e.g., `/ipfs/QmFoo`, or `/swarm/CID`, or `/dns/foobar.com`). This arbitrary string path can point to anything.

filips123:
@Alexander225

ZeroNet currently uses old style Bitcoin addresses format and Bitcoin cryptography.

But in the future, other formats and cryptographies may also be supported. So I would not like enforcing specific format here.

Based on this, namespace type would probably be the best. And for encoding/deciding it should then just use UTF8, just like `onion` in [content-hash](https://github.com/pldespaigne/content-hash). This would also make all future ZeroNet formats automatically supported.

alexander255:
@filips123: You don't need to specify a size with multicodec as the entire remainder of the string will be format-specific. As long as these future “formats & cryptographies” may be clearly distinguished by the application parsing the ZeroNet address from the current old-style Bitcoin address no new multicodec will be required, otherwise you'll likely need several multicodec values so that applications may distinguish which version of the ZeroNet address you're encoding (see `union` vs `union3`).

> And for encoding/deciding it should then just use UTF8, just like onion in content-hash. This would also make all future ZeroNet formats automatically supported.

Encoding/decoding is entirely your job with a multicodec, but you should prefer the most compact (binary) representation possible when transferring such data on the wire – otherwise the encoding format will likely not be approved as part of EIP1577. Don't even think about sending the human-readable/text version unless you absolutely cannot avoid it.

filips123:
@alexander255 I want this to support ZeroNet addresses as ENS content hashes. I will use previously linked `content-hash` library with almost same configuration as for `onion` and `onion3` domains. So no special encoding.

I need this PR and other PRs in language specific `multicodec` libraries because `content-hash` uses them to determine codec based on prefix.

So, based on this, which time do you think should be used? Probably namespace? Then I will fix this so this PR and others can be merged.

alexander255:
> I want this to support ZeroNet addresses as ENS content hashes. I will use previously linked content-hash library with almost same configuration as for onion and onion3 domains. So no special encoding.

I was talking about the (hex'd) binary encoding/decoding done in [content-hash/src/profiles.js](https://github.com/pldespaigne/content-hash/blob/master/src/profiles.js). What you get when you view the “human-readable” version of the ZeroNet address has to be encoded/decoded when storing as part of the multicodec value. This is not actually relevant for this discussion however, since all values stored in multicodec tend to be of some packed binary form.

> I need this PR and other PRs in language specific multicodec libraries because content-hash uses them to determine codec based on prefix.

Yes, and the reason they do that is because [EIP 1577](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1577.md#specification) mandates these value to be valid multicodec items.

> So, based on this, which time do you think should be used? Probably namespace? Then I will fix this so this PR and others can be merged.

Sorry for not really answering this earlier, unless I'm mistaken it would probably make sense to follow the example set by [Swarm](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1577.md#swarm).

Just to verify the bits of knowledge I've gathered along the way on this through:

  * An ZeroNet address is used too look up an `manifest.json` file using the DHT
  * This manifest file then describes the remainder of the website using some arbitrary linked-data format (ie: Using the manifest data I may be able to find the entire remaining content of the site using some kind of client software.)
  * The ZeroNet address is a SHA2-256 hash

If all of the above is correct I'd recommend using the same structure as Swarm, by reserving both an “ZeroNet namespace” and an “ZeroNet manifest IPLD resolver” multicodec value. Or in binary:

```
storage system: zeronet-ns (namespace, 0xee, ZeroNet path)
CID version: 1 (0x01)
content type: zeronet-manifest (ipld, 0xfd, ZeroNet manifest)
hash function: sha2-256 (0x12)
hash length: 32 bytes (0x20)
```

This way one could, for instance, properly display a ZeroNet page whose data has been retrieved over IPFS or Swarm. Please do tell me if I got something wrong!

Stebalien:
At a minimum, it looks like we'll need a namespace codec.

The remaining question is whether or not we need/want an IPLD codec:

* Can or will 0x ever support alternative key types?
* Can or will 0x ever support alternative hash functions?

If yes, we'd ideally format these names as `<0x-namespace><0x-key-cid>` to encode both the key type and hash function in the name (CIDs are `<cid-version><data-format><multihash>` tripples). (a

If no, we'd format these names as `<0x-namespace><raw-bitcoin-address>`.

filips123:
@alexander255 @Stebalien Are you sure all this is needed? ZeroNet uses completely custom network so I need multicodec just to use it in content hash. I already created PR for ZeroNet support in pldespaigne/content-hash#33 but it needs to wait until ZeroNet codec is released in libraries.

Stebalien:
We're trying to make everything as interoperable as possible. But, you're right. A "namespace" is the minimum and the rest is up to you.

filips123:
I updated codec to namespace type and reordered it because it was previously wrong.

filips123:
@Stebalien Can you now merge this PR?

filips123:
I also changed ZeroNet prefix to `0xe6` because almost all other `namespace` prefixes are in that range and `0xe6` is currently not used.

Stebalien:
All set.

PR #142, opened 2019-07-29
add codecs for ed25519 public keys and signatures with different hash types

### LINKED ISSUE #139: Add ed25519-sha2 signature code
Since it is possible to encode ed25519 (sha2?) public key already with `0xed`, it would be great to encode ed25519-sha2 signature in the same way. 

The following related issues are not actively updated, but the introduction of "multisig"/multisign may greatly help us in https://github.com/hyperledger/iroha to migrate from one cryptography algorithm to another without losing backward compatibility.
https://github.com/multiformats/multiformats/issues/23
https://github.com/multiformats/multicodec/pull/125

Ping @yusefnapora

alexmat2on:
Closes #139

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -70,6 +70,7 @@ ipfs-ns,                        namespace,      0xe3,           IPFS path
 swarm-ns,                       namespace,      0xe4,           Swarm path
 ipns-ns,                        namespace,      0xe5,           IPNS path
 ed25519-pub,                    key,            0xed,           Ed25519 public key
+ed25519-sig,                    signature,      0xef,           Ed25519 signature
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
 swarm-manifest,                 ipld,           0xfa,           Swarm Manifest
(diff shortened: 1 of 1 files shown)


yusefnapora:
@Stebalien is this all good?

Stebalien:
:thinking:

Stebalien:
@alexmat2on https://github.com/libp2p/specs/pull/217?

table.csv
```diff
@@ -70,6 +70,7 @@ ipfs-ns,                        namespace,      0xe3,           IPFS path
 swarm-ns,                       namespace,      0xe4,           Swarm path
 ipns-ns,                        namespace,      0xe5,           IPNS path
 ed25519-pub,                    key,            0xed,           Ed25519 public key
+ed25519-sig,                    signature,      0xef,           Ed25519 signature
```
cyphar: Why not make `ed25519-{pub,sig}` mean using Ed25519-SHA2 (since that's what effectively all Ed25519 users use), and then add only `ed25519-{pub,sig}-sha3`? It seems a bit odd to me to have two codec entries which are ambiguous.

PR #144, opened 2019-08-06
Add X25519 (Curve25519) pub key support

dmitrizagidulin:
Add support for X25519 public keys (for use with Key Agreement Keys etc).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -72,8 +72,8 @@ swarm-ns,                       namespace,      0xe4,           Swarm path
 ipns-ns,                        namespace,      0xe5,           IPNS path
 zeronet,                        namespace,      0xe6,           ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           Secp256k1 public key
-ed25519-pub,                    key,            0xed,           Ed25519 public key
 x25519-pub,                     key,            0xec,           Curve25519 public key
+ed25519-pub,                    key,            0xed,           Ed25519 public key
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
 swarm-manifest,                 ipld,           0xfa,           Swarm Manifest
--- a/table.csv
+++ b/table.csv
@@ -73,6 +73,7 @@ ipns-ns,                        namespace,      0xe5,           IPNS path
 zeronet,                        namespace,      0xe6,           ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           Secp256k1 public key
 ed25519-pub,                    key,            0xed,           Ed25519 public key
+x25519-pub,                     key,            0xec,           Curve25519 public key
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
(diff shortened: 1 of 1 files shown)


Stebalien:
(notification backlog, sorry)

Out of curiosity, how are (were?) you planning on using this?

Also note: this code is in the very limited one-byte range. Does it _really_ need to just be one byte or could you deal with a two-byte code (we have quite a few more of those).

dlongley:
> Out of curiosity, how are (were?) you planning on using this?

It's being used to express X25519 public keys, typically in DID (Decentralized Identifier) Documents, that can be used to derive secret keys for secure communications with an entity, typically identified by a DID. (see: https://github.com/digitalbazaar/x25519-key-pair and https://github.com/digitalbazaar/did-method-key-js).

> Also note: this code is in the very limited one-byte range. Does it really need to just be one byte or could you deal with a two-byte code (we have quite a few more of those).

Hmm, I now see:

> The first 127 bits are encoded as a single-byte varint, hence they are reserved for the most widely used multicodecs. So if you are adding your own codec to the table, you most likely would want to ask for a codec bigger than 0x80.

Originally we only saw the three step process for adding something to the table and glossed over this advice. We've been using ed25519 keys (directly related to x25519 keys) that use the memorable `0xed` code and reasoned that `0x25` looked like a good and memorable code for x25519 keys. Since ed25519 is in the class of codes that is bigger than `0x80`, I suppose we can explore picking something else if this is a sticking point.

@dmitrizagidulin -- can you take a look at this and our existing code to see how difficult it would be to change this (and perhaps recommend another code)?

dmitrizagidulin:
@dlongley No problem. It's not a bad change, it's just a couple places in one file in https://github.com/digitalbazaar/x25519-key-pair repo. (Plus the ripple of package version updates.)

As for suggesting a new code - I'm not fully clear on how the two-byte/varint encoding mechanism works, but let me take a look.

oed:
This would be great to have. @dmitrizagidulin maybe change the code to `0xe8` (right after `secp256k1-pub`?

dlongley:
@dmitrizagidulin, let's change this to 0xec (right next to 0xed which is for Edwards25519 and this is for Curve25519) to address the above issue and get the libraries patched.

dmitrizagidulin:
@dlongley Perfect, will do.

dmitrizagidulin:
@Stebalien I've changed the X25519/Curve25519 entry to `0xec`.

table.csv
```diff
@@ -73,6 +73,7 @@ ipns-ns,                        namespace,      0xe5,           IPNS path
 zeronet,                        namespace,      0xe6,           ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           Secp256k1 public key
 ed25519-pub,                    key,            0xed,           Ed25519 public key
+x25519-pub,                     key,            0xec,           Curve25519 public key
```
dlongley: Out-of-order fix:

```suggestion
x25519-pub,                     key,            0xec,           Curve25519 public key
ed25519-pub,                    key,            0xed,           Ed25519 public key
```

dmitrizagidulin: 🤦‍♂ riight, it's that kind of morning. Thanks!

Stebalien:
:tada:

PR #145, opened 2019-08-07
add a TLS multicodec

Stebalien:
Alternatively, we could repurpose https (443) as TLS and define HTTPS to be `/tls/http` (which is, IMO, more correct).

But let's make this decision quickly. I'm fine keeping this as it is and later deprecating `/https` in favor of `/tls/http`.


lidel:
While we are at it, would be good to quickly decide what should be the canonical representation of HTTPS, and ensure it is the default in the libraries we use internally for converting URLs to multiaddrs.

- JS: what should we use?  `/tls/http` vs `/https`  https://github.com/tableflip/uri-to-multiaddr/issues/6

PR #146, opened 2019-08-30
added "ident20" (truncated identity) and mut-* hash function for bloc…

michel47:
I would like to register 3 multihash codes supporting blockring addresses :
- ident20
- mut-160
- mut-224


Stebalien:
Can you post a link to something describing what these are?

Note: As far as I can tell, `ident20` is just `identity`.

PR #154, opened 2020-01-26
Add entry for BLAKE3 with 256-bit output (default)

nvzqz:
BLAKE3 is a fast cryptographic hash function. It is only one algorithm with no variants. See https://github.com/BLAKE3-team/BLAKE3 for details.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -15,6 +15,7 @@ keccak-224,                     multihash,      0x1a,           keccak has varia
 keccak-256,                     multihash,      0x1b,
 keccak-384,                     multihash,      0x1c,
 keccak-512,                     multihash,      0x1d,
+blake3,                         multihash,      0x1e,           BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
 dccp,                           multiaddr,      0x21,
 murmur3-128,                    multihash,      0x22,
 murmur3-32,                     multihash,      0x23,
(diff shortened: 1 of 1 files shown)


Stebalien:
* I've renamed it to blake3 given that the output _size_ is encoded in the multihash itself. We don't need per-output-size variants for blake3 because smaller outputs are prefixes of larger outputs.
* I've changed the code to a two byte code (encoded). No need to allocate this from the 3 byte range.

nvzqz:
8b8161834b3661641de5d6b9357e0c89f95c5af4 looks good to me. Good call on reducing the bytes needed to encode it. 👍

Stebalien:
@vmx or @Kubuxu thoughts? I considered moving this into the 1 byte range but I haven't seen any security analyses of the hash function yet. I'd rather not allocate a 1 byte code to something that might be broken (although that ship has sailed with definitely broken functions...).

Kubuxu:
I would say, allocate it one byte code. It is the same core function as blake2, the structure of the hash is different but with core staying the same, it should be secure.

Stebalien:
> different but with core staying the same, it should be secure

Yeah, you're probaly right. My concern is that they've reduced the number of rounds based on https://eprint.iacr.org/2019/1492.pdf and there hasn't been much time for anyone to refute those claims.

However, given the popularity of this new algorithm and the fact that nobody has come out and said "it's totally broken, don't use it", it's probably safe to allocate a 1 byte code (although I still don't plan on switching anything to it any time soon).

vmx:
I agree that BLAKE3 probably deserves a 1-byte code. Though I'm unsure if we really should do it right now. Is it an option to add it to the 2-byte code range and if it gets widely used assigning it a 1-byte one? This would waste one 2-byte slot and you would have two encodings for the same hash, but is that really an issue?

Stebalien:
@vmx we could, but if BLAKE3 becomes popular, we'd need to keep both codes and the code could get a bit tricky. If we think BLAKE3 will be popular, I'd like to just assign a 1 byte code out of the box.

PR #155, opened 2020-01-30
Remove out of date "sortable table" of prefixes

parkan:
The linked table at https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT is badly out of date and confused me for a good 15 minutes before I realized what was up. The normal table works fine, so let's just remove the outdated link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Multicodec is used in various [Multiformats](https://github.com/multiformats/mul
 
 ## Multicodec table
 
-Find the canonical table of multicodecs at [table.csv](/table.csv). There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
+Find the canonical table of multicodecs at [table.csv](/table.csv).
 
 ### Adding new multicodecs to the table
 


vmx:
@Stebalien It seems like to be an old version of https://github.com/Stebalien/static-table as the sorting doesn't really work. Can you please deploy a new version?

Stebalien:
The sorting definitely doesn't work for the "tag" field but it looks like it works for the other fields. I haven't made any newer versions so updating won't help, we'd have to fix the bug.

parkan:
sorry, was looking at this late at night and didn't realize what the root cause was

but now I see why I thought it was broken:

- sorting behavior is weird as @vmx said
- it somehow breaks search, so I couldn't find `0xe3`, for example, see below:

search

I think the latter may happen due to the dynamic row insertion that seems to happen (which also makes scrolling behave strangely)

overall I think this view is more confusing than the normal CSV in github and stand by my original suggestion

parkan:
code

@Stebalien  here's the table "sorted" by `code`, not sure what's going on here but it doesn't look like any ordering I know

Stebalien:
Yeah, I should have played with that a bit more before merging it.

Veronique-SideMeet:
When I follow the link: [https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv) it renders an empty page, saying that no data source was provided.

PR #157, opened 2020-01-31
add 0x0301: libp2p-peer-record mapping.

raulk:
It would be great to have the ability to reserve ranges in this table. I'd like to reserve the `[0x0300,0x03ff]` range for libp2p.

In the context of https://github.com/libp2p/go-libp2p/issues/776.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -100,6 +100,7 @@ p2p-websocket-star,             multiaddr,      0x01df,
 http,                           multiaddr,      0x01e0,
 json,                           serialization,  0x0200,         JSON (UTF-8-encoded)
 messagepack,                    serialization,  0x0201,         MessagePack
+libp2p-peer-record,             libp2p,         0x0301,         libp2p peer record type
 x11,                            multihash,      0x1100,
 sm3-256,                        multihash,      0x534d,
 blake2b-8,                      multihash,      0xb201,         Blake2b consists of 64 output lengths that give different hashes


yusefnapora:
Thanks @raulk! It would be great to get these reserved so we can stop using a string ID for peer records before we merge things in.

Stebalien:
> It would be great to have the ability to reserve ranges in this table. I'd like to reserve the [0x0300,0x03ff] range for libp2p.

We don't usually reserve ranges for specific applications:

1. Ranges don't really mean anything. We don't have an IPLD range, a Multiformats range, etc.
2. Different ranges encode to different sizes (when we use varints). 0-127 encodes to 1 byte, 128-16383 encodes to 2 bytes, etc.

What's the motivation for reserving this range?

raulk:
@Stebalien looking at the table as a passer-by, it seems there's already some contextual grouping, e.g.

* 0x01..: multiaddr (except for dag-json)
* 0x02..: serialisation formats.
* 0xb2..: blake.
* 0xb3..: skein.

Stebalien:
> it seems there's already some contextual grouping, e.g.

People do try to group codecs, sort of, but we don't explicitly _reserve_ ranges. I don't want to set a precedent where people come in and ask to reserve entire sections of, say, the 2 byte code range. For example, the request here would have reserved 1.5% of all two byte codes.

Note:

* We allocated individual blake/skein codes as a block because we needed one code per length.
* We allocated most of the serialization formats and multiaddr codes together at the same time.
* Many of the 1 & 2 byte codes were allocated before we started really thinking about running out of space (md5 etc. should not be in those ranges).

PR #160, opened 2020-02-03
add support for secp256k1-pub

awoie:
The decentralized identity community is using DIDs as their foundation layer, i.e., implementing a decentralized PKI system. did:key is one specific DID method and is using multibase/multicodec to express the method-specific-identifier of their DIDs. Currently, there is only support for ed25519-pub but the DID community uses secp256k1 also extensively. In order to use secp256k1 with did:key, we need to get support for secp256k1-pub.


OR13:
+1, any blockers I can help with?

imachug:
In what format are point coordinates stored? OpenSSL? Raw? Custom format?

PR #164, opened 2020-03-15
Update table.csv

Morpheusai:
Add codec for taucoin


vmx:
@Morpheusai you need quite a feed codecs and we have limited space on the smaller varint numbers. Could you please move it to a two bytes range (similar to the Holochain codec)?

PR #165, opened 2020-03-16
Feat/draft v standard

Stebalien:


mikeal (dismissed):
We need some way of noting that this is a breaking change for any dependent projects.

Especially since this column is not being appended to the end, this is going to change/break the representation that comes back from many CSV parsers. It’s pretty common to throw out the header line and just parse each line into an array rather than parsing each line into a map using the header values as key names.

Stebalien:
I agree, that's why the tests are failing. But I'm not sure how to actually _do_ that.

mikeal:
Is anyone pulling the latest master automatically or are all the known libraries keeping their own cached version?

Stebalien:
go-multihash is pulling into a submodule but (a) that's locked against a specific commit and (b) we're only using it for tests.

@vmx?

README.md
```diff
@@ -35,13 +35,24 @@ Multicodec is used in various [Multiformats](https://github.com/multiformats/mul
 
 Find the canonical table of multicodecs at [table.csv](/table.csv). There's also a sortable [viewer](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv).
 
+### Status
+
+Each multicodec is marked with a status:
+
+* draft - this codec has been reserved but may be reassigned if it doesn't gain wide adoption.
+* standard - this codec has been widely adopted and may not reassigned.
```
vmx: What about renaming "standard" to "permanent" like the [URI Schemes Registry](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) is doing it. That would make clear that having a standard, might still not be enough to get the permanent assignment here.

I think "draft" is a good name.

vmx:
In the JS world there is a script that takes the file that generates JS code. It's a manual process.
In Rust, there is not even a script, there it's totally manually.

rvagg:
So, merging this and breaking people's stuff will tell them that they should probably be relying on the header to know what the columns are rather than hard-wiring it in. So I'm +1 on merging and taking care of the consumers that we control and adding in documentation that if you're consuming the table _you must use the header to determine column contents_.

ribasushi:
You know - we could just switch to a slightly more extensible format: JSON, YAML, whatever: something that allows us to grow the metadata without rebreaking the world every time.

We can also keep the .csv around by siply having a step that regenerates it from the now-canonical master-spec.

I would volunteer to implement this, together with the necessary CI hooks, provided there is buy in to go that route.

rvagg:
XML please, lashings of it, with namespaces too, and a schema while you're at it!

+1 to a JSON version as the primary. We could even just publish it to npm as is for consumption by js-multicodec and js-multihash (maybe). Let's see what @vmx says.

vmx:
For this case I don't like JSON for the lack of being able to add comments, but so is CSV => I think we should go with JSON as canonical representation, and generate CSV out of it (as CSV is really so much easier to read/look things up).

mikeal:
Why don’t we just use a markdown table and write a simple script that bulids a .csv and .json file on every checking using GitHub Actions?

rvagg:
> just use a markdown table

Because it's not that much different from a CSV then, we haven't solved very much. We may as well have an "input CSV" an a "consume-me CSV" if we head down that route. The only thing a Markdown table really buys us is the ability to embed it in a large Markdown doc (the table is getting large enough that this probably isn't something we need to be doing) or to discourage it from being consumed as the primary source (documentation can do that job, as can naming).

@vmx is there much more than we'd need in terms of comments than we could fit into a `"comments"` field for each entry? That can be arbitrary text with comments about anything relating to an entry.

vmx:
> @vmx is there much more than we'd need in terms of comments than we could fit into a `"comments"` field for each entry?

Good point. I think that's enough.

Stebalien:
I don't have the time to refactor how we store multicodecs but I really want this merged so we can more easily merge new multicodec registrations.

table.csv
```diff
@@ -1,462 +1,462 @@
...
+dash-block,                     ipld,           0xf0,           draft,     Dash Block
+dash-tx,                        ipld,           0xf1,           draft,     Dash Tx
+swarm-manifest,                 ipld,           0xfa,           draft,     Swarm Manifest
+swarm-feed,                     ipld,           0xfb,           draft,     Swarm Feed
+udp,                            multiaddr,      0x0111,         draft,
+p2p-webrtc-star,                multiaddr,      0x0113,         draft,
+p2p-webrtc-direct,              multiaddr,      0x0114,         draft,
+p2p-stardust,                   multiaddr,      0x0115,         draft,
+p2p-circuit,                    multiaddr,      0x0122,         permanent,
+dag-json,                       ipld,           0x0129,         draft,     MerkleDAG json
```
vmx, line 96: I would consider dag-json being permanent. But we can also discuss this in a seperate PR.

table.csv
```diff
@@ -1,462 +1,462 @@
...
+udt,                            multiaddr,      0x012d,         draft,
+utp,                            multiaddr,      0x012e,         draft,
+unix,                           multiaddr,      0x0190,         permanent,
+thread,                         multiaddr,      0x0196,         draft,     Textile Thread
+p2p,                            multiaddr,      0x01a5,         permanent, libp2p
+ipfs,                           multiaddr,      0x01a5,         permanent, libp2p (deprecated)
```
vmx, line 102: Would it make sense to call `ipfs` draft and `p2p` to be permanent?

table.csv
```diff
@@ -1,462 +1,462 @@
...
+blake2b-448,                    multihash,      0xb238,         permanent,
+blake2b-456,                    multihash,      0xb239,         permanent,
+blake2b-464,                    multihash,      0xb23a,         permanent,
+blake2b-472,                    multihash,      0xb23b,         permanent,
+blake2b-480,                    multihash,      0xb23c,         permanent,
+blake2b-488,                    multihash,      0xb23d,         permanent,
+blake2b-496,                    multihash,      0xb23e,         permanent,
+blake2b-504,                    multihash,      0xb23f,         permanent,
+blake2b-512,                    multihash,      0xb240,         permanent,
+blake2s-8,                      multihash,      0xb241,         draft,     Blake2s consists of 32 output lengths that give different hashes
```
vmx, line 195: Is intentionally Blake2**b** permanent and Blake2**s** draft?

vmx:
@Stebalien you mention that having "draft" and "permanent" would make merging new multicodecs easier. My impression is that this wasn't really the problem so far when we didn't merge things. I think the problem is that we miss a clear definition what multicodec codes (and their categories) are about. It boils down to: Is it just a random table with numbers in it, or do they have a certain meaning.

table.csv
Stebalien: Good point.

table.csv
Stebalien: I'm not aware of anyone using the blake2s codec in practice.

Stebalien:
> It boils down to: Is it just a random table with numbers in it, or do they have a certain meaning.

Well, _that_ part usually boils down to "are they using IPLD right". But draft v permanent would make me feel better about adding new codes.

vmx (approved):
Before merging I'd like to also have the approval from @rvagg, as he also spent quite a lot of time on all this.

table.csv
rvagg: you could also make this argument for most of the blake2b hashes, `blake2b-256` is the only one in _wide_ use (because of Filecoin)

rvagg (approved):
I'm fine with the column, its the first-pass of filling it in that I think has issues. We can resolve them later but these are the most important discrepancies from my perspective:

* only one of the blake2b hashes has widespread usage yet they are all permanent here
* sha2-256-trunc254-padded and poseidon-bls12_381-a2-fc1 are the CID pairs for fil-commitment-unsealed and fil-commitment-sealed respectively. The former are draft but the latter are permanent.

Second-level concerns I have are:

* I think the bitcoin* and eth* codecs should be permanent, they are being used in the wild, outside of PL, and as more than just examples (dbl-sha2-256 would come along with bitcoin*)
* cbor and json would be nice to have as permanent, although I can't point to active non-example uses of them
* dag-cose is getting more use in the wild I think, but we could wait for input from those users

Stebalien:
> only one of the blake2b hashes has widespread usage yet they are all permanent here

My concern is that poking holes in this range could be troublesome if someone actually _needs_ one of the others. But then again, we might as well stick to our policy.

> sha2-256-trunc254-padded and poseidon-bls12_381-a2-fc1 are the CID pairs for fil-commitment-unsealed and fil-commitment-sealed respectively. The former are draft but the latter are permanent.

Will fix.

> I think the bitcoin* and eth* codecs should be permanent, they are being used in the wild, outside of PL, and as more than just examples (dbl-sha2-256 would come along with bitcoin*)

Fair enough.

> cbor and json would be nice to have as permanent, although I can't point to active non-example uses of them
> dag-cose is getting more use in the wild I think, but we could wait for input from those users

SGTM.

Stebalien:
> dag-cose is getting more use in the wild I think, but we could wait for input from those users

Jose or Cose or both?

Stebalien:
@rvagg I've fixed everything but jose/cose.

rvagg (approved):
leave cose/jose as draft for now

👍

PR #166, opened 2020-03-20
Add dag-jose, -cose

oed:
See discussion here: https://github.com/ipld/specs/issues/251

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -39,6 +39,8 @@ bencode,                        serialization,  0x63,           bencode
 dag-pb,                         ipld,           0x70,           MerkleDAG protobuf
 dag-cbor,                       ipld,           0x71,           MerkleDAG cbor
 libp2p-key,                     ipld,           0x72,           Libp2p Public Key
+dag-jose,                       ipld,           0x73,           MerkleDAG JOSE
+dag-cose,                       ipld,           0x74,           MerkleDAG COSE
 git-raw,                        ipld,           0x78,           Raw Git object
 torrent-info,                   ipld,           0x7b,           Torrent file info field (bencoded)
 torrent-file,                   ipld,           0x7c,           Torrent file (bencoded)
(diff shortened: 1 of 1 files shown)


vmx:
@oed Could you please use a code > 0x80. Codes less that that are encoded as 1-byte varints and as there is limited space, we'd like to keep them for widely used formats.

oed:
Sure I can change it @vmx. I'm wondering though what defines "widely used formats"? Right now the plan is for 3Box and Textile to use these codecs for our dag structures.

vmx:
> I'm wondering though what defines "widely used formats"?

That's indeed a good question. We need to get more specific about that. Currently we try to push as much as possible outside the 0x80 range (historically sadly there things in that range that now wouldn't).

rvagg:
Do we need to see implementations and/or specifications before accepting something new into the table? Seems like there's a risk in having people stake out territory for things that don't exist yet. Maybe leaving an open PR like this while something is formalised is an appropriate course of action - flagging that these are wanted but waiting for something concrete to prove they exist outside of this table.

vmx:
> Do we need to see implementations and/or specifications before accepting something new into the table?

We don't, but soon we'll have a new column to indicate the state of things: https://github.com/multiformats/multicodec/pull/165

I'll work on this soon.

mikeal:
I do worry a bit about adding this particular codec before we’ve written a spec, but I’m not going to block the PR.

Ideally, our libraries would be much more tolerant of prototyping stuff like this using an unregistered entry so that we can start building and writing specs before committing an entry to the table. But we aren’t there yet, so lets get this in.

It’s pretty uncontroversial that we need a `dag-cose` codec since it uses another tag we currently forbid in `dag-cbor`.

oed:
Fyi, the js implementation of `dag-jose` is planned to start next week :)

mikeal:
> Fyi, the js implementation of dag-jose is planned to start next week :)

Awesome! Once you have a PoC it would be great to get a spec for it in the [specs repo](https://github.com/ipld/specs), similar to what we have for [other `dag-` codecs](https://github.com/ipld/specs/tree/master/block-layer/codecs).

vmx:
@Stebalien any objections to merge this one? I think it's good to go ("MerkleDAG" is weird, but that's also what the other codecs are currently using, I'd change that in a separate PR).

vmx:
Finally. Sorry that merging this one took so long.

PR #173, opened 2020-05-01
feat: add BLS12-381 public keys

tplooker:
Adds the two variations of BLS12-381 public keys.


rvagg:
Actually, if we're accepting the `_` in #171 to make BLS12-381 clearer then I'd suggest the names be `bls12_381-g1-pub` and `bls12_381-g2-pub`. @vmx?

OR13:
+1 have plans to use this asap :)

tplooker:
@rvagg are we able to merge this?

rvagg:
Yeah, but I don't have access to do it. Will chase someone down or get access myself to do it.

rvagg:
👌

PR #175, opened 2020-05-15
Add ripemd definitions

mvandeberg:
As proposed in #117, there are no definitions for the ripemd family of hashes.

The previous PR would have been sufficient with the given feedback, but the author never made the requested changes and the PR is now over a year old.

This PR would replace #117.

It conforms to feedback on #117, with the more common ripemd-160 hash using the one byte space and the less common algorithms using the two byte space.

0x52 was chosen somewhat arbitrarily as the ASCII encoding of 'R'.

Furthermore, 0x51 is the serialization code for CBOR, which means 0x52 is unlikely to be needed as a value in relation to CBOR.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -32,7 +32,6 @@ dns6,                           multiaddr,      0x37,
 dnsaddr,                        multiaddr,      0x38,
 protobuf,                       serialization,  0x50,           Protocol Buffers
 cbor,                           serialization,  0x51,           CBOR
-ripemd-160,                     multihash,      0x52,
 raw,                            ipld,           0x55,           raw binary
 dbl-sha2-256,                   multihash,      0x56,
 rlp,                            serialization,  0x60,           recursive length prefix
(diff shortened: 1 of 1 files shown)


ribasushi:
Note that there are multiple precedents for "non-FIPS" hash functions to live outside of the single-byte space, the most notable example being `blake2b-256`. 

Is burning `0x52` for a function exclusively used in bitcoin addressing these days really warranted?

rvagg:
I'm kind of with @ribasushi on this and would rather keep them together as a family - it certainly makes it easier to look them up - but that does somewhat contradict the previous ask in #117.

What would be the justification for using one-byte for 160? Is there a common usage that we'd expect this to attract that would benefit from shortest-possible representation? Even in Bitcoin I think the only usage is over the top of a sha2-256 as well, a combination that might warrant its own entry in here (in the same fashion as dbl-sha2-256) if they were to be put into common usage.

Also there's the 0xbX block that's got plenty of holes if we're looking for an arbitrary spot in the one-byte range, `0xbf` might be easier to justify than the one here _if_ it's true that Bitcoin are pretty much the only users of 160 today.

vmx:
These days we try to put as little as possible in the <0x80 range. Some things in there are due to historic reasons and wouldn't likely end up there today. So I suggest moving the ripemd-160 also to the same range as the others.

mvandeberg:
ripemd-160 has been moved to the two byte range and is contiguous with other ripemd sizes.

PR #177, opened 2020-05-27
Add 0xcert asset imprints

fulldecent:
Adds an additional well-known format to the table.

0xcert Asset Imprints. This is a Merkle tree structure which is used to hash and identify arbitrary-depth objects (generally from JSON). This is an open specification.

Following is some documentation for reference and your consideration of this pull request:

- Overview: https://github.com/0xcert/framework/blob/master/packages/0xcert-cert/README.md
- Documentation: https://docs.0xcert.org/api/guides/update-asset-imprint.html#usage-overview
- Reference implementation: https://github.com/0xcert/framework/tree/master/packages/0xcert-cert

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -437,6 +437,7 @@ skein1024-1016,                 multihash,      0xb3df,
 skein1024-1024,                 multihash,      0xb3e0,
 poseidon-bls12_381-a2-fc1,      multihash,      0xb401,         Poseidon using BLS12-381 and arity of 2 with Filecoin parameters
 poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         Poseidon using BLS12-381 and arity of 2 with Filecoin parameters - high-security variant
+0xcert-imprint-256,             0xcert,         0xce11,         0xcert Asset Imprint (root hash)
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -437,6 +437,7 @@ skein1024-1016,                 multihash,      0xb3df,
 skein1024-1024,                 multihash,      0xb3e0,
 poseidon-bls12_381-a2-fc1,      multihash,      0xb401,         Poseidon using BLS12-381 and arity of 2 with Filecoin parameters
 poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         Poseidon using BLS12-381 and arity of 2 with Filecoin parameters - high-security variant
+0xcert-imprint-256,             0xcert,         0xce11,         0xcert Asset Imprint (root hash)
```
Stebalien, line 440: Wouldn't this count as an IPLD format?

fulldecent: It is a distinct specification.

Stebalien: IPLD is a more of a "meta" spec for interpreting arbitrary merkle-linked data. The bitcoin, eth, etc. codecs are considered IPLD codecs for this reason.

But we can always just merge this with it's own group and change it later when/if someone writes an IPLD format implementation for 0xcert.

Issue #181, opened 2020-07-19
Consider Changing 0xcert-imprint-256 Identifier As It Is Not A Valid Variable Name In Some Languages


bonedaddy:
I'm working on a multicodec implementation in C, and the `0xcert-imprint-256` name can't be used. Regardless of whether or not you are trying to define multicodecs in C via variables or macros, the compiler will report an error. This is likely to be the case in other C based languages as well. It's also a pretty poor choice of name, as in programming languages `0x` is used to indicate a hexadecimal value.

Consider changing to `zeroxcert-imprint-256` at the very least, or use some other name, otherwise this multicodec can't be included in any multicodec library based in a language that doesnt allow `0x` to be in variable names.

PR #182, opened 2020-07-20
Naming restrictions (doc & validate), rename `0xcert` to be code-friendly

rvagg:
Ref: https://github.com/multiformats/multicodec/issues/181
Ref: https://github.com/multiformats/multicodec/pull/177

/cc @bonedaddy @fulldecent 

I agree with @bonedaddy on this one, it's awkward for `0xcert` but they kind of chose an intentionally awkward name, I bet their code has issues with constant naming.

@fulldecent can you live with this rename or does this violate something fundamental about `0xcert`?


bonedaddy:
`zeroxcert-imprint-256` seems good to me, and means I'll be able to include it in the C multicodec library I'm writing :D

fulldecent:
Yes, that's fine. It's just an identifier.

PR #185, opened 2020-07-31
Add BLS 12-381 g1 and g2 public key concatenated identifier

tplooker:
For certain signatures such as https://github.com/mattrglobal/bbs-signatures it is useful to have a single private key have a corresponding public key in both the G1 and G2 fields.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -79,6 +79,7 @@ bls12_381-g1-pub,               key,            0xea,           BLS12-381 public
 bls12_381-g2-pub,               key,            0xeb,           BLS12-381 public key in the G2 field
 x25519-pub,                     key,            0xec,           Curve25519 public key
 ed25519-pub,                    key,            0xed,           Ed25519 public key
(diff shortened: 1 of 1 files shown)


OR13:
Looking forward to using this representation.

OR13:
@rvagg @vmx can we get a review of this? let us know if we need to make any adjustments... the reason we need an encoding for both keys concatenated is to support https://github.com/w3c-ccg/ldp-bbs2020/issues/37

vmx:
@OR13 could you please put it after `ed25519-pub`, we currently sort by code value (so that it's easier to spot if some value is already taken).

tplooker:
@vmx done

vmx:
> we probably should have put `bls12_381*` in the two-byte range

It is in the 2-byte range. Everything > `0x80` is in the two-byte range.

vmx (approved):

PR #186, opened 2020-08-02
Add P-384

OR13:
AFAIK, this will be the first NIST Curve added here... so its pretty valuable for legacy systems.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -79,6 +79,7 @@ bls12_381-g1-pub,               key,            0xea,           BLS12-381 public
 bls12_381-g2-pub,               key,            0xeb,           BLS12-381 public key in the G2 field
 x25519-pub,                     key,            0xec,           Curve25519 public key
 ed25519-pub,                    key,            0xed,           Ed25519 public key
+p384-pub,                       key,            0xef,           P-384 public key
 dash-block,                     ipld,           0xf0,           Dash Block
 dash-tx,                        ipld,           0xf1,           Dash Tx
 swarm-manifest,                 ipld,           0xfa,           Swarm Manifest


rvagg:
This is fine by me, but we're now out of space in `0xeX`, I'm wondering if we should just start populating a new space for curve public keys, since there's potentially a lot of them, leaving the current ones in place. There's 4 more prime curves in the NIST set that would go along with just this one and there's no space to nicely order them if we lock this one in.

Thoughts from anyone else?

@OR13 just out of interest, do you have a current use-case for a P-384 public key?

OR13:
yes, the main problem is that web crypto / hardware backed kms often don't support secp256k1, ed25519 or bls12381... so what do you do if you are required to work with say ... a customer who requires you use NIST curves for somethings ?

I would like to be able to generate a non extractable software isolated key in a browser, and represent it using `did:key`... 

https://github.com/w3c-ccg/did-method-key

I have done so for ed25519, secp256k1, and bls12381... but I have no way of doing it for a NIST curve without registering at least 1. I chose P-384 because of its use by both the US and NZ governments.

[PDF Download of GCSB  Approved Cryptographic Algorithms ](https://www.nzism.gcsb.govt.nz/pdf/index/2106)

OR13:
@rvagg yes, its probably better to register all the NIST Curves nicely in their own block... I was wondering how we were going to go higher than `ef` :) I'm not attached to the prefix length at all.

OR13:
if someone can recommend the block to register for NIST curves, I will close this PR and open open one which registers them.

rvagg:
Anything above the single byte range is wide open, I see two slots that might be nice (this is mostly a subjective call given the space available!). After `x11`, we could probably start a block in `0x12XX` for common curve pubkeys. There's also space we could work on in 0xb5XX, the preceding poseidon entries are somewhat related (in that they _involve_ a curve).

OR13:
@rvagg 

`0x12XX` seems like the right place for these: https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve

0x1200 - P-256
0x1201 - P-384
0x1202 - P-521
0x1203 - Ed448
0x1204 - X448

^ would this be correct?...

OR13:
Closing this PR in favor of https://github.com/multiformats/multicodec/pull/190

Issue #49, opened 2017-06-21
Make CID versions multicodecs


Stebalien:
This way, we can trivially disambiguate CIDs from, e.g., raw multihashes.

(To do this, we'd need to steal 0x1 from unary for CIDv1).

Stebalien:
@diasdavid and I discussed this offline but I figured it should be posted here to keep things public.

We and came across the theoretical problem of handling non-multibase CIDv18 (18 = 0x12 = SHA2-256 multicodec) and decided that we could just skip that or define CIDv0 = CIDv18.

hobofan:
Is there a way to get this moving? Are there any blockers that need to be discussed?

I'm currently in the process of drafting our new format for [rlay-project/rlay-ontology](https://github.com/rlay-project/rlay-ontology), which strongly revolves around CIDs.
In our current (older) format, we always have to have some knowledge which fields of a struct have which types (CID or multicodec), and for the fields that accept both, we have a not so pretty "try to parse as CID and if that fails it's a multicodec" logic. Adding CIDs as proper multicodecs would be great step towards simplifying this.

rvagg:
I think the next step to move this forward is to open a PR against table.csv to add something like `cid, ipld, 0x1, CIDv1`, you're welcome to do that @hobofan and maybe we can move forward. The step after this might be to implement the spec change detailed @ https://github.com/multiformats/cid/issues/25 to cement it.

`0x1` isn't in here anymore, I believe it was extracted when multibase was split out and along the way somewhere `base1` / unary was dropped entirely, so it's up for grabs here without needing to do conflict resolution.

I'm still not clear on why #94 is necessarily a different thing though tbh. It seems to be part of the same bundle of concerns and just a matter of definitions.

Stebalien:
#94 is proposing adding a _new_ multicodec for CIDs. So a multicodec-identified CID would be `<cid-multicodec><v1><ipld-codec>...`. This proposal is to just treat `v1` as a multicodec (`<v1><ipld-codec> -> <cid-multicodec><ipld-codec>...`).

oed:
Hey @Stebalien @rvagg I'm happy to create the PRs needed to help move this forward, as I'm interested in using multicodec similarly to CIDs without it causing conflicts. Is there anything I need to be aware of in terms of the two different options? (to me your approach @Stebalien makes the most sense)

Stebalien:
I believe all we need is two PRs:

[...]

PR #187, opened 2020-08-07
Define CIDv1 as a mutlicodec

oed:
This PR adds CIDv1 as a multicodec with value `0x01`. 

Fixes https://github.com/multiformats/multicodec/issues/49

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -1,6 +1,8 @@
 name,                           tag,            code,           description
 identity,                       multihash,      0x00,           raw binary
 cidv1,                          ipld,           0x01,           CIDv1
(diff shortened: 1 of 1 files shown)


mikeal:
Good idea.

We should reserve 2 and 3 as well for future versions.

oed:
Sure @mikeal can update the PR!

Stebalien:
> We should reserve 2 and 3 as well for future versions.

It can't hurt to do that, but I don't think it's _critical_ as we can always pick a different ID. But then again it would be nice to have them be sequential.

PR #189, opened 2020-08-12
Add Ceramic Document Id

oed:
Adds an entry for Ceramic Document Ids, as `0xce`. I think this would be a `namespace`, but let me know if it should be something different!

See how it's used: https://github.com/ceramicnetwork/CIP/issues/59

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -65,6 +65,7 @@ bitcoin-tx,                     ipld,           0xb1,           Bitcoin Tx
 bitcoin-witness-commitment,     ipld,           0xb2,           Bitcoin Witness Commitment
 zcash-block,                    ipld,           0xc0,           Zcash Block
 zcash-tx,                       ipld,           0xc1,           Zcash Tx
+docid,                          namespace,      0xce,           Ceramic Document Id
 stellar-block,                  ipld,           0xd0,           Stellar Block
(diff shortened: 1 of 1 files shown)


oed:
Ping, any blockers here?

vmx (approved):
Thanks for pinging, it slipped my attention.

Sounds good to me, I'd like to have one more approval from @rvagg or @Stebalien before merging.

rvagg:
@oed would it be a problem for this to be in the two-byte range rather than taking up another slot in the quickly-filling-up single byte range? We're trying to push newer entries down in the table unless there's good reason for something to be that short, because of the space premium at the top.

oed:
@rvagg Every update made to a ceramic document needs to contain the document id currently. Trimming one byte of the DocID would definitely be preferable!

vmx:
It is already in the two-byte range as it is > `0x80`. The bytes in the list are the raw identifiers, and not the varints.

oed:
Right, my bad. I guess I'd be fine with this.

rvagg (approved):

vmx:
@oed Please rebase once more and I'll merge it right after.

oed:
@vmx done!

Issue #188, opened 2020-08-07
Add KangarooTwelve hash?


tmladek:
Hello,
I couldn't find the KangarooTwelve hash variant (https://keccak.team/kangarootwelve.html) from the Keccak family, is there a reason it is not included? (Seeing as the rest of the Keccak functions are). Or am I misunderstanding, and it actually does correspond to some of the already included variants?

The reason it's interesting is because it's significantly more performant than the rest of the Keccak functions, and even recommended for such purposes by the Keccak team itself: https://public-inbox.org/git/91a34c5b-7844-3db2-cf29-411df5bcf886@noekeon.org/

Would you consider adding it? Thanks!

Stebalien:
There's no reason not to include it but the major multihash implementations aren't likely to include an implementation of the hash function by default to keep code size down. If you're interested, file a PR and pick a multicodec in the 2-3 byte range. See https://github.com/multiformats/multicodec#adding-new-multicodecs-to-the-table.

OT: Given ever increasing hardware support, I'd generally just suggest sha256 unless you have an issue with length extension attacks.

PR #192, opened 2020-08-25
add / reserve KangarooTwelve in table.csv

tmladek:
Submitting PR as per issue #188, thanks in advance for your consideration.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -122,6 +122,7 @@ p384-pub,                       key,            0x1201,         P-384 public Key
 p521-pub,                       key,            0x1202,         P-521 public Key
 ed448-pub,                      key,            0x1203,         Ed448 public Key
 x448-pub,                       key,            0x1204,         X448 public Key
+kangarootwelve,                 multihash,      0x1d01,         fast and secure extendable-output hash function from the Keccak family
 sm3-256,                        multihash,      0x534d,
 blake2b-8,                      multihash,      0xb201,         Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,
--- a/table.csv
+++ b/table.csv
@@ -122,7 +122,7 @@ p384-pub,                       key,            0x1201,         P-384 public Key
 p521-pub,                       key,            0x1202,         P-521 public Key
 ed448-pub,                      key,            0x1203,         Ed448 public Key
 x448-pub,                       key,            0x1204,         X448 public Key
-kangarootwelve,                 multihash,      0x1d01,         fast and secure extendable-output hash function from the Keccak family
+kangarootwelve,                 multihash,      0x1d01,         KangarooTwelve is an extendable-output hash function based on Keccak-p
 sm3-256,                        multihash,      0x534d,
 blake2b-8,                      multihash,      0xb201,         Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,


rvagg:
Unfortunately I think this one is going to be more complicated than than a single entry since it's an extendible output function, so we're going to need a mess similar to what we have for blake2b and skein right now.

I'm not sure about others' feelings about this but I'm not super keen to see another massive block for a little used hash function. Maybe we could tackle this by soft-reserving a block with enough space to fit in the multiple-of-8 entries like blake2b and skein and just adding in likely common ones to start with, maybe 32, 64 and 128, and leaving the holes for others to be added if ever anyone shows up wanting them?

Anyone else?

tmladek:
Oh, I thought that's what the second varint specifying the digest length was for!

(blake2b having a "mess" because the variants differed on block length, not the output length)

In that case I'd agree, and would be happy with what you're proposing

-------- Original Message --------

rvagg:
Well .. it's true that you can specify the length in a multihash and that's one way of handling variable length outputs, and we have `blake3` as a precedent in here for that, although it's at least got a strong 32-byte output default (recommendation). afaik KangarooTwelve doesn't have anything similar, it's an entirely different hash depending on the output length you use, not just a truncation, and there's no default you could guess at if you were wanting to verify validity of length.

@Stebalien, @vmx, thoughts?

vmx:
> it's an entirely different hash depending on the output length you use

Nope, it's the same hash, just truncated.

```console
vmx@gene$ python3 K12.py 'hello world!' 8
c571e7ff9fb12b76
vmx@gene$ python3 K12.py 'hello world!' 16
c571e7ff9fb12b760f0fb9b02ea06f15
vmx@gene$ python3 K12.py 'hello world!' 32
c571e7ff9fb12b760f0fb9b02ea06f15b3e3970855b1210c01b404ad2d7609e3
```

I've modified the [reference implementation](https://github.com/XKCP/XKCP/blob/66121748b9fbff3641d51374d117e349c3db94f3/Standalone/KangarooTwelve/Python/K12.py) and added to the bottom:

```python
if __name__ == '__main__':
    import binascii
    import sys

    inputMessage = sys.argv[1].encode('utf-8')
    outputByteLen = int(sys.argv[2])
    digest = KangarooTwelve(inputMessage, b'', outputByteLen)
    print(binascii.hexlify(digest).decode('ascii'))
````

File: table.csv [hunk 1]
```diff
@@ -122,6 +122,7 @@ p384-pub,                       key,            0x1201,         P-384 public Key
 p521-pub,                       key,            0x1202,         P-521 public Key
 ed448-pub,                      key,            0x1203,         Ed448 public Key
 x448-pub,                       key,            0x1204,         X448 public Key
+kangarootwelve,                 multihash,      0x1d01,         fast and secure extendable-output hash function from the Keccak family
```
rvagg, line 125: OK moving on then with this, a minor nit, aside from capitalisation, my suggestion for the description here:

> KangarooTwelve is an extendable-output hash function based on Keccak-p

tmladek: Understood, I just wanted to mention why it makes sense to include in the first place (the performance), but am in no way insisting upon keeping it.

And what about capitalisation? None of the names in the `name` column are capitalised, and the `description`s don't seem to start with a capital letter either?

rvagg: Just suggesting the change for the `description` field (starting with "KangarootTwelve ..."). Most of the descriptions (where they exist), it's not totally consistent but let's go with the majority; it's particularly apt when naming something that's been capitalised by the author though so that's easy.

rvagg (approved):

tmladek:
Thank you kindly.

PR #195, opened 2020-09-08
Add RSA public key (x.509 Encoded)

b5:
Add a prefix for raw (_not_ libp2p) x.509 public key bytes. I think we should have this for a parity with other public key formats.

It's worth adding _now_ because it'll make writing code that does key rotation from RSA -> ED25519 with open standards possible.

Regarding the `0x5d` prefix There's a W3C working group [draft spec](https://github.com/w3c-ccg/did-method-key) floating out in the wild that would require this addition to make RSA keys work, and [this issue](https://github.com/w3c-ccg/lds-ed25519-2018/issues/3) makes reference to `0x5d` as the chosen prefix. I've reused this value here to be just slightly above picking at random.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -37,6 +37,7 @@ protobuf,                       serialization,  0x50,           Protocol Buffers
 cbor,                           serialization,  0x51,           CBOR
 raw,                            ipld,           0x55,           raw binary
 dbl-sha2-256,                   multihash,      0x56,
+rsa-pub,                        key,            0x5d,           RSA public key (X.509 encoded)
 rlp,                            serialization,  0x60,           recursive length prefix
 bencode,                        serialization,  0x63,           bencode
 dag-pb,                         ipld,           0x70,           MerkleDAG protobuf


rvagg:
I think we're trying to avoid assigning single-byte entries as much as possible. Perhaps this might rise above the bar, although I don't really have a feel for where that bar might be and how close this comes to it. @Stebalien @vmx, thoughts?

vmx:
To me the bar for single byte entries is. Would it be a problem to put it into the 2-bytes (or more) range? If the answer is "why not", then definitely put it into that other range.

@b5 Is there a reason why it needs to be in the single byte range, or could it well be in some other range?

jonnycrunch:
@jonnycrunch just adding myself to the discussion as a reminder that is cool stuff @b5 and I need to understand it more!

b5:
No reason at all it needs to be in the single byte range. Given the use case is mostly around migration, I agree it makes more sense to use the two byte (or more range). I mainly kicked this PR off to get this discussion going, and would welcome a byte assignment from spec maintainers.

At the risk of being presumptuous, I think RSA is starting to lose favourability in circles that will be interested in multiformats, (even in our case we're interested in this multicodec so we can transition to Edwards curve keys), so I fully agree it should move to the 2 bytes or more range.

If others don't have a suggestion in the next day or two I'll pick something in the 2 byte range to keep the conversation moving, but I'd much prefer to be told what works here :)

rvagg:
We have a block in `0x1200` that _was_ going to be curve keys, but we could just as easily say that `0x1200` is `-pub` keys and `0x1300` is `-priv` keys, so this would probably fit there. They're all `key` anyway. 🤷

b5:
superseded by #226

PR #198, opened 2020-09-21
Add Skynet codecs

peterjanbrone:
This PR adds a Skynet codec to ultimately add ENS support for Skylinks.  
Skynet is a decentralized CDN built on top of the Sia network, find out more [here](https://siasky.net/).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -458,3 +458,4 @@ holochain-key-v0,               holochain,      0x947124,       Holochain v0 pub
 holochain-key-v1,               holochain,      0x957124,       Holochain v1 public key + 8 R-S (63 x Base-32)
 holochain-sig-v0,               holochain,      0xa27124,       Holochain v0 signature  + 8 R-S (63 x Base-32)
 holochain-sig-v1,               holochain,      0xa37124,       Holochain v1 signature  + 8 R-S (63 x Base-32)
+skynet-skylink-v1,              skynet,         0xd101,         Skynet v1 skylink (Base64) 
--- a/table.csv
+++ b/table.csv
@@ -458,5 +458,6 @@ holochain-key-v0,               holochain,      0x947124,       Holochain v0 pub
 holochain-key-v1,               holochain,      0x957124,       Holochain v1 public key + 8 R-S (63 x Base-32)
 holochain-sig-v0,               holochain,      0xa27124,       Holochain v0 signature  + 8 R-S (63 x Base-32)
 holochain-sig-v1,               holochain,      0xa37124,       Holochain v1 signature  + 8 R-S (63 x Base-32)
-skynet-ns,                      namespace,      0xd101,         Skynet Namespace
-skynet-skylink-v1,              skynet,         0xd102,         Skynet v1 skylink (Base64)
+sia-ns,                         namespace,      0x1991,         Sia Namespace
+skynet-ns,                      namespace,      0x1992,         Skynet Namespace
+skynet-skylink-v1,              skynet,         0x1993,         Skynet v1 skylink (Base64)
--- a/table.csv
+++ b/table.csv
@@ -458,6 +458,5 @@ holochain-key-v0,               holochain,      0x947124,       Holochain v0 pub
 holochain-key-v1,               holochain,      0x957124,       Holochain v1 public key + 8 R-S (63 x Base-32)
 holochain-sig-v0,               holochain,      0xa27124,       Holochain v0 signature  + 8 R-S (63 x Base-32)
 holochain-sig-v1,               holochain,      0xa37124,       Holochain v1 signature  + 8 R-S (63 x Base-32)
-sia-ns,                         namespace,      0x1991,         Sia Namespace
-skynet-ns,                      namespace,      0x1992,         Skynet Namespace
-skynet-skylink-v1,              skynet,         0x1993,         Skynet v1 skylink (Base64)
+skynet-ns,                      namespace,      0xb19910,         Skynet Namespace
+skynet-skylink-v1,              skynet,         0xb19911,         Skynet v1 skylink (Base64)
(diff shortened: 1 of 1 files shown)


peterjanbrone:
@vmx @rvagg @Stebalien, I apologize for pinging you directly, but would it be possible for one of you to take a look at this PR? I am trying to add ENS support to Skynet and having a codec is a requirement ^^

Stebalien:
Could you write up a very brief summary for what each one of these is? It helps to ensure they're categorized correctly. Specifically:

* What's the practical difference between the "skynet" namespace and the "sia" namespace? I.e., how will they be used.
* What is skynet-skylink-v1 (practically)? E.g., are you going to have links of the form `<skynet-ns><skynet-skylink-v1>ENCODED_LINK`.
* How important is _space_. Everything below `0x4000` is in the two byte range (we have ~16K available codecs in this range), everything above that is in the 3 byte range (where we have ~2M codecs). For example, if something is always going to be base64 encoded, I assume space isn't too much of an issue.

Other: This table is sorted by codec. Could you move this section up?

Off topic: I noticed you're hard-coding base64 for skylink links. I'd like to suggest that you use multibase (https://github.com/multiformats/multibase/), you'll save yourself some pain down the road when you, e.g., try to integrate with browsers.

peterjanbrone:
@Stebalien Thank you for getting back to me. 

- I only added `sia-ns` to reserve the codec, as we have no use for it currently I have removed it
- I need the codec to be able to create a CID, for that I believe I only need `skynet-skylink-v1`, and the links thus won't be of the form  `<skynet-ns><skynet-skylink-v1>ENCODED_LINK`. I modeled this much as `swarm` did and they have defined a namespace so ideally I would like to keep `skynet-ns` as namespace if possible. 

`skynet-skylink-v1` is an identifier for what we call a Skylink, it is very comparable to an IPFS hash, and we do plan to upgrade to v2 at some point, to extend the schema or make things possible that we have currently not foreseen

- I have updated the codec to be in the 3byte range.

Thank you for the tip on https://github.com/multiformats/multibase/ - we will definitely look into it, we probably can't make that change though as we do not want to exceed 46 bytes at all costs and we have already optimised to get it down to 46 as is.

aschmahmann:
@peterjanbrone

> Multibase is a protocol for disambiguating the encoding of base-encoded (e.g., base32, base36, base64, base58, etc.) binary appearing in text.

This means that your size in the binary/wire representation will not increase by using multibase only the text representation will.

peterjanbrone:
> @peterjanbrone
> 
> > Multibase is a protocol for disambiguating the encoding of base-encoded (e.g., base32, base36, base64, base58, etc.) binary appearing in text.
> 
> This means that your size in the binary/wire representation will not increase by using multibase only the text representation will.

I see, thank you, I will definitely look into it! I do fear it's not something we are likely to adopt any time soon seeing as it would change the textual representation of a Skylink. We need that to remain the same for the time being.

aschmahmann:
> I do fear it's not something we are likely to adopt any time soon seeing as it would change the textual representation of a Skylink. We need that to remain the same for the time being.

That's totally understandable concern, changing user facing identifiers is hard. IPFS actually experienced similar issues around the introduction of CID (i.e. [CIDv0](https://github.com/multiformats/cid#cidv0) has implicit assumptions like the base encoding and the codec that were needed to make existing tooling compatible with the new CID format).

However, I'm a little confused. You've said you'd like to use this codec in a CID and the CID (on the wire and text) is going to be different from how you're working with existing Skylinks (on the wire and text) now (e.g. the newly reserved codec number will show up in the CID) how are you planning to keep things the same? Additionally the CIDv1 spec comes with multibase included so I'm not sure I understand why you think it wouldn't be compatible.

One more thing. I'm not an expert on ENS by any means, but if you're just looking for ENS compatibility their spec https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1577.md seems to imply that just having the namespace is enough for things to work. For example just having `<skylink-ns><skylink-address>` should be sufficient.

Arachnid:
Correct - you don't have to use CIDs if they're not appropriate; you just need to obtain a multicodec namespace ID, and encode the content hash field as `<skylink-ns><skylink-address>`, specifying this format somewhere (such as in an EIP). For compactness it'd be preferable if the address was in binary form rather than text format for compactness.

peterjanbrone:
I have dropped all codecs but one, being the `skynet-ns`.

peterjanbrone:
@vmx @rvagg @Stebalien, if you would be so kind to give this another review please. We would like to move ahead and integrate ENS as soon as possible.

table.csv
```diff
@@ -458,3 +458,4 @@ holochain-key-v0,               holochain,      0x947124,       Holochain v0 pub
 holochain-key-v1,               holochain,      0x957124,       Holochain v1 public key + 8 R-S (63 x Base-32)
 holochain-sig-v0,               holochain,      0xa27124,       Holochain v0 signature  + 8 R-S (63 x Base-32)
 holochain-sig-v1,               holochain,      0xa37124,       Holochain v1 signature  + 8 R-S (63 x Base-32)
+skynet-ns,                      namespace,      0xb19910,         Skynet Namespace
```
rvagg, line 461: minor nit - you have two extra spaces before the comment that should be removed

rvagg:
I will defer to @aschmahmann & @Stebalien on this one as they understand the domain better

aschmahmann:
@peterjanbrone @rvagg yep, the 3 byte namespace LGTM :+1: (after the small table formatting issue above is resolved)

-----

@peterjanbrone as an aside if you're interested (now or in the future) in getting Skynet to utilize CIDs I'd be happy to discuss it with you, just give me a ping :smile:

aschmahmann (approved):

table.csv
rvagg: let's get this resolved and then this can be merged @peterjanbrone

peterjanbrone:
@rvagg updated, thanks!

PR #201, opened 2020-10-28
Adds Textile Thread multiaddr

carsonfarmer:
The thread protocol is now relatively well-recognized in the community. Perhaps its time to make it official?

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -97,6 +97,7 @@ dag-json,                       ipld,           0x0129,         MerkleDAG json
 udt,                            multiaddr,      0x012d,
 utp,                            multiaddr,      0x012e,
 unix,                           multiaddr,      0x0190,
+thread,                         multiaddr,      0x0196,         Textile Thread
 p2p,                            multiaddr,      0x01a5,         libp2p
 ipfs,                           multiaddr,      0x01a5,         libp2p (deprecated)
 https,                          multiaddr,      0x01bb,


sbiaudet (approved):

sbiaudet (approved):

Stebalien:
What is a thread multiaddr used to address? Usually, multiaddrs are used to address endpoints/machines, not resources/content.

However, this issue comes up so frequently I'm writing a proposal to see if we can unify around a "multipath" concept: https://github.com/multiformats/multiformats/pull/55.

carsonfarmer:
Yup, multiformats/multiformats#55 is probably _exactly_ what we're looking for. In theory, a thread multiaddr is used to address a resource available via a particular peer/endpoint. But it _does_ go one step further than "endpoint/machine" to address a resource.

carsonfarmer:
I see the conversation around the above linked "multipath" proposal is somewhat stalled. What is the likelihood of moving forward on that, versus just working with the code proposed in this PR?

rvagg:
@Stebalien you're probably the authority for that question ^, specifically does:

> But it does go one step further than "endpoint/machine" to address a resource.

raise `thread` to being something we can fit in here?

Stebalien:
Yes, that's why I made the multipath proposal. It's a simple generalization of multiaddr.

rvagg (approved):

Issue #205, opened 2021-02-07
missing several variations of sha2


warpfork:
The multiformats table seems to be missing several variations of sha2.  Specifically:

- sha2-224
- sha2-384
- sha2-512-224
- sha2-512-256

See the excellent comparison chart in Wikipedia for more details here: https://en.wikipedia.org/wiki/SHA-2#Comparison_of_SHA_functions

Or I've lifted just the most relevant rows in a screenshot here for ease of reference:

![Comparison of SHA2 functions](https://user-images.githubusercontent.com/627638/107147008-e75c9480-694b-11eb-8f22-12cd1d085150.png)

I'm *especially* interested in getting sha2-384 into the table, because it's both widely available and offers some innate defenses against length extension attacks (unlike the sha2 variants we currently have in the table, which both offer none!).  But all four of these are well-standardized and deserve an indicator number in the multiformats table.

---

Naming notes:

- Most documents seem to refer to these as e.g. "sha-224" rather than "sha2-224".  However, most of the multiformat table seems to prefix things with the family name, so I suppose we should say "sha2-224" to continue that pattern.
- Most documents seem to refer to the latter two as e.g. "sha-512/224" (with a slash).  However, most of the multiformat table seems to prefer dashes for separations, so I suppose we should say "sha2-512-224" to continue (both) patterns.

---

Language library availability notes:

In golang, all of these are available in the standard library (as `sha256.New224`, `sha512.New384`, `sha512.New512_224`, and `sha512.New512_256` respectively).

In javascript, in the browser, taking the [Mozilla docs for SubtleCrypto.digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) as a reference, SHA-384 is widely available.

---

I'll file a PR for this soon, but first, a couple **questions**:

- Four new numbers are needed for this.  Is it preferred to look for the first range where I can find four new consecutive numbers?  Or should I stuff two in the first open range, and the other two later?
- Am I doing the naming conventions correctly?

warpfork:
Some additional clarifications that @aschmahmann pointed out:

- _Are these just truncations of the other existing functions?_
  - **No**.  These are not trivially constructable in most scenarios.  (You cannot take a black-box of a sha2-512 function and -- taking just the output, without reaching into the hash function's guts -- chop it down into a sha2-384 output.  There's more that varies than just the output size.)

[...]

PR #206, opened 2021-02-15
Add missing variations of sha2

warpfork:
Per discussion in https://github.com/multiformats/multicodec/issues/205 .

I've inserted these entries at the first available numbers following the other sha2 variants.  It's unfortunate we can't put them all in a row, but no space is available since some sha3 variants immediately follow the already-entabled sha2 variants.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -19,9 +19,13 @@ keccak-256,                     multihash,      0x1b,
 keccak-384,                     multihash,      0x1c,
 keccak-512,                     multihash,      0x1d,
 blake3,                         multihash,      0x1e,           BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
+sha2-224,                       multihash,      0x1f,           aka SHA-224; specified by FIPS 180-4.
+sha2-384,                       multihash,      0x20,           aka SHA-384; specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,
 murmur3-128,                    multihash,      0x22,
 murmur3-32,                     multihash,      0x23,
+sha2-512-224,                   multihash,      0x24,           aka SHA-512/224; specified by FIPS 180-4.
+sha2-512-256,                   multihash,      0x25,           aka SHA-512/256; specified by FIPS 180-4.
 ip6,                            multiaddr,      0x29,
 ip6zone,                        multiaddr,      0x2a,
 path,                           namespace,      0x2f,           Namespace for string paths. Corresponds to `/` in ASCII.


vmx:
I would move `sha2-512-224` and `sha2-512-256` to higher ranges. But I don't have a strong opinion on that.

table.csv
```diff
@@ -19,9 +19,13 @@ keccak-256,                     multihash,      0x1b,
 keccak-384,                     multihash,      0x1c,
 keccak-512,                     multihash,      0x1d,
 blake3,                         multihash,      0x1e,           BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
+sha2-224,                       multihash,      0x1f,           aka SHA-224; specified by FIPS 180-4.
+sha2-384,                       multihash,      0x20,           aka SHA-384; specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,
 murmur3-128,                    multihash,      0x22,
 murmur3-32,                     multihash,      0x23,
+sha2-512-224,                   multihash,      0x24,           aka SHA-512/224; specified by FIPS 180-4.
```
rvagg, line 27: IMO these should be up in the `0x1012` range. The single byte range is precious. `sha2-224` is unlikely to get wide usage too so I'd prefer that to be higher as well. I can see the case for 384 though.

fwiw I don't think FIPS 180-4 is a strong claim to their level of importance at this stage, in crypto terms it's ancient, if you're making new software then it's not highly likely that you're reaching for it as your standard. And FIPS-202 is all about SHA3 I believe.

warpfork: Oh I'm definitely not claiming a FIPS number expresses anything about their importance :) That's just there to remove any possible ambiguity about the naming vs what the thing is.

mvdan:
I also don't have a strong opinion, but I have to admit I'd avoid taking spots in the single-byte range unless the added multicodec will be *very* common :)

vmx:
I also lean towards putting it into higher ranges (as @rvagg suggested) and it seems most people involved lean towards higher ranges, lets put it into higher ranges and be done. I guess that most important thing is that it is somewhere.

warpfork:
I'm pretty strongly held that I want sha2-384 to be in the single-byte range because it is the "best" out of these hashes (see discussion in issue) that is also available in the standard libraries we tend to notice first (js, in node and browsers, and golang).

I guess I'll bump up the `sha2-512-*` numbers to two-byte if we like.

I could also see bumping up `sha2-224` to a higher range because I don't think it'll be commonly used.  Fragments things that are related, but I guess avoiding that isn't really a goal we need to prioritize, so okay.

warpfork:
... I seem to have started this PR using the github edit feature, and then deleted the fork it creates in the meanwhile, so I no longer know how to amend this PR.  Going to close it and open a new one shortly.

PR #207, opened 2021-02-16
Add SoftWare Heritage persistent IDentifier snapshots

Ericson2314:
See
https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html
for the technical details of that spec.

Software Heritage has done a superb job promoting content addressing in
general, and their identifier scheme (SWHIDs, for short) in particular.
By supporting them in CIDs / IPLD, I hope the IPFS ecosystem can align
itself with that effort.

Per the linked documentation, SWHIDs have their own nested grammar and
versioning scheme. I begun by taking the version 1 core identifier
grammar, unrolled it, and replaced `:` with `-` per the guidelines on
separators, with the result being these 5 rows.

There is overlap between the remaining for and git-raw, so this just
adds SWHID snapshots for now.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -115,6 +115,7 @@ ws,                             multiaddr,      0x01dd,         permanent,
 wss,                            multiaddr,      0x01de,         permanent,
 p2p-websocket-star,             multiaddr,      0x01df,         permanent,
 http,                           multiaddr,      0x01e0,         draft,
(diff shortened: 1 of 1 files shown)


vmx (approved):
Before merging I'd like to get an approval/thumbs up from other maintainers as well.

warpfork (approved):
I'm good with this.  :+1: 

I think no one has raised any objections to this code, in particular, at all -- so I'll merge this shortly if no objections appear.

PR #210, opened 2021-05-08
Add secp256k1 private key

OR13:
Reasoning:

There already exists and ed25519 priv key: https://github.com/multiformats/multicodec/blob/master/table.csv#L128

Without the ability to represent both public and private components of asymmetric key pairs users will be forced for do things like this:

```
'id': '',
'type': 'EcdsaSecp256k1VerificationKey2020',
'controller': 'did:example:123',
'publicKeyMultibase': 'zQ3shU9mxTYfMTLceuppoNM5siowHvw3xBk21r855kwgNHNAX',
'privateKeyBase58': '6GyEn649EsbEEJXGQ5m8Eq6NANDmHTPQPWGyF5uL7kx8'
```

This prevents efficient compression using multibase codecs.

We will likely need to add entries for NIST curves and BLS12381 as well.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -79,6 +79,7 @@ swarm-ns,                       namespace,      0xe4,           draft,     Swarm
 ipns-ns,                        namespace,      0xe5,           draft,     IPNS path
 zeronet,                        namespace,      0xe6,           draft,     ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           draft,     Secp256k1 public key
+secp256k1-priv,                 key,            0xe8,           draft,     Secp256k1 private key
 bls12_381-g1-pub,               key,            0xea,           draft,     BLS12-381 public key in the G1 field
 bls12_381-g2-pub,               key,            0xeb,           draft,     BLS12-381 public key in the G2 field
 x25519-pub,                     key,            0xec,           draft,     Curve25519 public key
--- a/table.csv
+++ b/table.csv
@@ -79,7 +79,7 @@ swarm-ns,                       namespace,      0xe4,           draft,     Swarm
 ipns-ns,                        namespace,      0xe5,           draft,     IPNS path
 zeronet,                        namespace,      0xe6,           draft,     ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           draft,     Secp256k1 public key
-secp256k1-priv,                 key,            0x1301,           draft,     Secp256k1 private key
+secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
 bls12_381-g1-pub,               key,            0xea,           draft,     BLS12-381 public key in the G1 field
 bls12_381-g2-pub,               key,            0xeb,           draft,     BLS12-381 public key in the G2 field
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -79,6 +79,7 @@ swarm-ns,                       namespace,      0xe4,           draft,     Swarm
 ipns-ns,                        namespace,      0xe5,           draft,     IPNS path
 zeronet,                        namespace,      0xe6,           draft,     ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           draft,     Secp256k1 public key
+secp256k1-priv,                 key,            0xe8,           draft,     Secp256k1 private key
```
OR13, line 82:
```suggestion
secp256k1-priv,                 key,            0x1301,           draft,     Secp256k1 private key
```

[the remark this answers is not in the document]
OR13: implemented in https://github.com/multiformats/multicodec/pull/210/commits/016319e1cf9a1a26269ffc6b4ed8ff86c05066a9

OR13:
@rvagg thanks for the review! 

@msporny and I have been having this argument for years :)

Its not as much of an issue with JSON-LD when you can just pretend that not defining the term is somehow making it safer to to do this:


```
'id': '',
'type': 'EcdsaSecp256k1VerificationKey2020',
'controller': 'did:example:123',
'publicKeyBase58': 'zQ3shU9mxTYfMTLceuppoNM5siowHvw3xBk21r855kwgNHNAX',
'privateKeyBase58': '6GyEn649EsbEEJXGQ5m8Eq6NANDmHTPQPWGyF5uL7kx8'
```

The "problem" with multibase... is that you can't not define the entry... and still use it to represent a private key... like you can with JSON-LD.

Folks using JSON-LD with `privateKeyBase58` or `privateKeyMultibase` can decide if omitting context definitions is a security feature or not... but certainly nobody can even create a `privateKeyMultibase` without a registry entry here... 

the alternative to defining `privateKeyMultibase` is to use `privateKeyJwk`, which already preserves key type information and public key representations, and relies on IANA... in order for multibase to ever improve on JWK, it must be at least as expressive, imo... its a show stopper for our adoption of it.. if we can't represent privateKeyMultibase for the key types we use, we will just keep using privateKeyJwk... I expect most developers will do the same when they realize they cannot completely represent a private key.

table.csv
```diff
@@ -79,6 +79,7 @@ swarm-ns,                       namespace,      0xe4,           draft,     Swarm
 ipns-ns,                        namespace,      0xe5,           draft,     IPNS path
 zeronet,                        namespace,      0xe6,           draft,     ZeroNet site address
 secp256k1-pub,                  key,            0xe7,           draft,     Secp256k1 public key
+secp256k1-priv,                 key,            0x1301,           draft,     Secp256k1 private key
```
rvagg, line 82:
```suggestion
secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
```
sorry, two stray chars now

rvagg (changes requested):

rvagg (approved):

PR #214, opened 2021-05-23
Clarify public key encodings

oed:
The affected curves can be represented in both a compressed and an uncompressed format. This clarifies that the compressed format should be used.


rvagg (approved):
👌

awoie (approved):
lgtm

PR #215, opened 2021-05-26
Make TLS permanent

carpawell:
There are https://github.com/multiformats/go-multiaddr/pull/153 and https://github.com/multiformats/multiaddr/pull/109 PR's that add TLS.

Also, TLS is used in examples in README in the [main repo](https://github.com/multiformats/multiaddr/blame/master/README.md#L121).

Therefore, it seems logical to make it permanent.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -106,7 +106,7 @@ onion,                          multiaddr,      0x01bc,         draft,
 onion3,                         multiaddr,      0x01bd,         draft,
 garlic64,                       multiaddr,      0x01be,         draft,     I2P base64 (raw public key)
 garlic32,                       multiaddr,      0x01bf,         draft,     I2P base32 (hashed public key or encoded public key/checksum+optional secret)
-tls,                            multiaddr,      0x01c0,         draft,
+tls,                            multiaddr,      0x01c0,         permanent, Transport Layer Security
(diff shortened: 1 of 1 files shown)


rvagg:
> permanent - this codec has been widely adopted and may not reassigned.

I would say that while TLS is widely adopted itself, that use of the codec code here may not be - and you've linked to PRs, not merged and released code.

The fact that it's `draft` shouldn't be interpreted as it being at imminent risk of being removed, that's not the case at all. We've removed very few entries from the table and mostly that's happened immediately after merge when we realised it was a mistake. So it's more of a journey _toward_ permanent, with the preference to not lock in entries where they aren't in heavy use - like `dag-pb`, for instance.

So I'm -1 on this for now, but it's certainly a good candidate for permanent if we see this code being widely used.

carpawell:
So it is ok, if `draft` != `can be deleted`. My only misunderstanding was about using it as an example in the main repo and also that `TLS` is certainly a wildly used protocol, but somehow it is "draft".

BTW, both PRs that I linked were merged.

Stebalien:
> My only misunderstanding was about using it as an example in the main repo and also that TLS is certainly a wildly used protocol, but somehow it is "draft".

TLS is widely used. The TLS _multicodec_ is not.

PR #219, opened 2021-06-17
Add Curve25519 private key.

msporny:


rvagg (approved):
* we have the pubkey @ `0xec`
* we've previously agreed (and continued to have discussions about) private keys in the table
so this seems fine to me

msporny:
> * we've previously agreed (and continued to have discussions about) private keys in the table
>   so this seems fine to me

Note that there already is a private key in the table -- ed25519-priv

OR13:
Would be nice to register support for NIST curves as well.

OR13:
Related PRs...

https://github.com/multiformats/multicodec/pull/194

https://github.com/multiformats/multicodec/pull/210

PR #221, opened 2021-07-01
add a Noise codepoint

marten-seemann:
Analogous to https://github.com/multiformats/go-multiaddr/pull/153, we also want to be able express Noise addresses, e.g. `/ip4/127.0.0.1/tcp/4001/noise`.

cc @mxinden

PR #222, opened 2021-07-02
add a codepoint for plaintextv2

marten-seemann:
Analogous to #221.

As we want to stop negotiating security protocols, we'll also need a codepoint for insecure connections. Example: `/ip4/1.2.3.4/tcp/4001/insecure`.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -469,6 +469,7 @@ poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,     Posei
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
 fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
+plaintextv2,                    multiaddr,      0x706c61,       draft,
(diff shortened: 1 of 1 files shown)


Stebalien:
Do we? This is for testing only and we'll need to support security transport negotiation for the foreseeable future anyways.

Stebalien (approved):
Ok, so this is a real protocol and it does make sense to support it. But let's put it in the three byte range.

table.csv
```diff
@@ -108,6 +108,7 @@ garlic64,                       multiaddr,      0x01be,         draft,     I2P b
 garlic32,                       multiaddr,      0x01bf,         draft,     I2P base32 (hashed public key or encoded public key/checksum+optional secret)
 tls,                            multiaddr,      0x01c0,         draft,
 noise,                          multiaddr,      0x01c6,         draft,
+insecure,                       multiaddr,      0x01cb,         draft,
```
mxinden, line 111: When dialing a Multiaddr with the `insecure` protocol, which security protocol would be used to exchange the peer IDs? [Plaintext v2](https://github.com/libp2p/specs/tree/master/plaintext)? Why not define this as the `plaintextv2` protocol?

```suggestion
plaintextv2,                       multiaddr,      0x01cb,         draft,
```

mxinden (dismissed):

table.csv
Stebalien: Yes, plaintextv2. And yeah, that might be a better name.

marten-seemann:
I renamed the protocol to `plaintextv2`, and assigned a 3-byte code-point (PLA in ASCII, sorry for not being more creative).

Stebalien (approved):

mxinden (commented):
:+1: thanks!

PR #226, opened 2021-08-29
Add RSA public key

expede:
Add RSA public key. Updated over https://github.com/multiformats/multicodec/pull/195

Changed from #195:
* Move to `0x12XX` range
* Added X.509 directly to the name (happy to take out)

RSA is has length variants, I suppose that we're going to use multiformats to specify key length. Alternately, we could specify rsa-2048-pub, rsa-4096-pub, and so on, but it feels orthogonal _to me_.

Thoughts or changes @b5?

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -132,6 +132,7 @@ p384-pub,                       key,            0x1201,         draft,     P-384
 p521-pub,                       key,            0x1202,         draft,     P-521 public Key (compressed)
 ed448-pub,                      key,            0x1203,         draft,     Ed448 public Key
 x448-pub,                       key,            0x1204,         draft,     X448 public Key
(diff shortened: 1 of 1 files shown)


b5:
nope looks good to me! I'll close the other PR in favour of this one

expede:
Oh, since this is a new PR, I should probably ping @rvagg 👋 No rush, just want to make sure that you'll see it!

rvagg (approved):

rvagg:
This all seems good to me.

I'm not sure the lengths are orthogonal, but if, for practical purposes, you don't need that signalling up front when consuming the bytes then maybe it's unnecessary? If it turns out to be needed then we could easily add additional entries and maybe make this one a common length, 2048 perhaps. Or even rename it if we're not too far down that road. Will look to those who are using it to provide feedback on the utility of the length differentiation.

PR #229, opened 2021-09-20
add multiaddr for p2p-circuit-inner

marten-seemann:
As defined in https://github.com/libp2p/specs/pull/353.


Stebalien (approved):

mxinden (approved):
Looks good to me.

As pointed out in https://github.com/libp2p/specs/pull/359 I chose `p2p-circuit-inner` due to a lack of a better name. Happy for more intuitive suggestions.

vyzo (commented):
we need a better name for this.

PR #234, opened 2021-09-30
Add missing variations of sha2

warpfork:
Per discussion in https://github.com/multiformats/multicodec/issues/205 ,
and further discussion of the numbers,
in https://github.com/multiformats/multicodec/pull/206 .

Most of the variants are shifted into larger number ranges.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -19,6 +19,7 @@ keccak-256,                     multihash,      0x1b,           draft,
 keccak-384,                     multihash,      0x1c,           draft,
 keccak-512,                     multihash,      0x1d,           draft,
 blake3,                         multihash,      0x1e,           draft,     BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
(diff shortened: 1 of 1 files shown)


vmx (approved):
I'd like to wait from an approval from @rvagg before this gets merged.

I don't know if sha2-384 is widely used so that it should be in the 1 byte range (we got stricter about putting things in the 1 byte range, so looking into existing hashes there isn't really meaningful). Though I also don't see a point of blocking it if there's consensus that it is good idea.

warpfork:
I will absolutely fight about it, forever.  As I have said repeatedly, but will say yet one more time: it is the only hash that is in the standard libraries we especially care about, and is also some amount of strong against length-extension attacks.  I will use it and I will recommend its use constantly based on these factors.

rvagg (approved):
yeah, this seems reasonable to me, 384 is common enough, the others either aren't, or maybe shouldn't be

mvdan (approved):
please merge before we forget about it again :)

PR #236, opened 2021-10-14
update murmur3 length / family to match current uses

willscott:
0x22 is currently used widely through IPFS as representing the `murmur3_64` length. This makes the table match reality, and allocates 0x24 for the 128 bit length.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -21,7 +21,7 @@ keccak-512,                     multihash,      0x1d,           draft,
 blake3,                         multihash,      0x1e,           draft,     BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
 sha2-384,                       multihash,      0x20,           permanent, aka SHA-384; as specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,           draft,
-murmur3-64,                     multihash,      0x22,           draft,     The murmur3 variant used for UnixFS directory sharding.
+murmur3-x64-64,                 multihash,      0x22,           draft,     The first 64-bits of a murmur3-x64-128, used for UnixFS directory sharding.
 murmur3-32,                     multihash,      0x23,           draft,
 ip6,                            multiaddr,      0x29,           permanent,
 ip6zone,                        multiaddr,      0x2a,           draft,
@@ -132,7 +132,7 @@ sha2-256-trunc254-padded,       multihash,      0x1012,         permanent, SHA2-
 sha2-224,                       multihash,      0x1013,         permanent, aka SHA-224; as specified by FIPS 180-4.
 sha2-512-224,                   multihash,      0x1014,         permanent, aka SHA-512/224; as specified by FIPS 180-4.
 sha2-512-256,                   multihash,      0x1015,         permanent, aka SHA-512/256; as specified by FIPS 180-4.
-murmur3-128,                    multihash,      0x1022,         draft,
+murmur3-x64-128,                multihash,      0x1022,         draft,
 ripemd-128,                     multihash,      0x1052,         draft,
 ripemd-160,                     multihash,      0x1053,         draft,
 ripemd-256,                     multihash,      0x1054,         draft,
--- a/table.csv
+++ b/table.csv
@@ -21,9 +21,8 @@ keccak-512,                     multihash,      0x1d,           draft,
 blake3,                         multihash,      0x1e,           draft,     BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
 sha2-384,                       multihash,      0x20,           permanent, aka SHA-384; as specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,           draft,
-murmur3-64,                     multihash,      0x22,           draft,
+murmur3-64,                     multihash,      0x22,           draft,     The murmur3 variant used for UnixFS directory sharding.
 murmur3-32,                     multihash,      0x23,           draft,
-murmur3-128,                    multihash,      0x24,           draft,
 ip6,                            multiaddr,      0x29,           permanent,
 ip6zone,                        multiaddr,      0x2a,           draft,
 path,                           namespace,      0x2f,           permanent, Namespace for string paths. Corresponds to `/` in ASCII.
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -21,8 +21,9 @@ keccak-512,                     multihash,      0x1d,           draft,
 blake3,                         multihash,      0x1e,           draft,     BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
 sha2-384,                       multihash,      0x20,           permanent, aka SHA-384; as specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,           draft,
-murmur3-128,                    multihash,      0x22,           draft,
+murmur3-64,                     multihash,      0x22,           draft,
```
aschmahmann, line 24: We should document in the description exactly what we mean by this. What we mean is the version of murmur3 used by UnixFS as that's been the first (and only?) user of this code AFAICT.

table.csv
```diff
@@ -21,8 +21,9 @@ keccak-512,                     multihash,      0x1d,           draft,
...
 murmur3-32,                     multihash,      0x23,           draft,
+murmur3-128,                    multihash,      0x24,           draft,
```
aschmahmann, line 26: Do people use this? Should it be higher up in the table? I'm willing to wait on a user here asking for this before allocating a codec.

Also, murmur3-128 is not a defined thing. There is a 32bit and 64bit version of it. So we'd need a codec for each of the ones we care about. (e.g. murmur3-128-x86 and murmur3-128-x64)

aschmahmann (changes requested):

table.csv
rvagg: there isn't a 64-bit version of murmur3 btw, it's usually a truncation of 128 or some other abomination

murmur3 has x86 and x64 versions of 32-bit and 128-bit output lengths so when we invoke "64" we're doing weird things, but that's just the nature of murmurhash unfortunately!

if we need a truncated form because we're using it, then maybe we rename the entry to be clear, `murmur3-x64-128-64` or `murmur3-x64-128-trunc64` ("truncX" is a pattern we've used in the table already but for an entry that zeros out some bits, not actually shrinks the output length proper, but overloading the term should be OK).

table.csv
rvagg: I think you need to look at those as `0x22` which has "murmur3-128" here already. Let's just fix up `0x22` to have a better name, or fix up the uses of it such that it's clear what kind of sorcery is being invoked and not add new entries. Nobody is happy with murmur3 in this table anymore so lets not expand it, at least here in the 1 byte range.

table.csv
```diff
@@ -132,6 +132,7 @@ sha2-256-trunc254-padded,       multihash,      0x1012,         permanent, SHA2-
 sha2-224,                       multihash,      0x1013,         permanent, aka SHA-224; as specified by FIPS 180-4.
 sha2-512-224,                   multihash,      0x1014,         permanent, aka SHA-512/224; as specified by FIPS 180-4.
 sha2-512-256,                   multihash,      0x1015,         permanent, aka SHA-512/256; as specified by FIPS 180-4.
+murmur3-128,                    multihash,      0x1022,         draft,
```
rvagg, line 135:
```suggestion
murmur3-x64-128,                multihash,      0x1022,         draft,
```

rvagg (approved):
lgtm, but I'd like to leave this open for enough time to let eyes that normally see these to weigh in since this is a thorny one

warpfork (approved):
LGTM.

I definitely approve of the choice to change the name of the 0x22 code.

I'm indifferent to what we change the name to -- and trust that the discussion here covered that already, so :+1: .

Agree with @rvagg that I don't even see it as important to add a code for murmur3-128.  But if we do, I also don't object to something in a multibyte range (and the PR now has it as 0x1022, so, alright, fine by me).  Good by me to merge with that; also good by me to drop that new entry and merge.

table.csv
```diff
@@ -21,7 +21,7 @@ keccak-512,                     multihash,      0x1d,           draft,
 blake3,                         multihash,      0x1e,           draft,     BLAKE3 has a default 32 byte output length. The maximum length is (2^64)-1 bytes.
 sha2-384,                       multihash,      0x20,           permanent, aka SHA-384; as specified by FIPS 180-4.
 dccp,                           multiaddr,      0x21,           draft,
-murmur3-128,                    multihash,      0x22,           draft,
+murmur3-x64-64,                 multihash,      0x22,           permanent, The first 64-bits of a murmur3-x64-128 - used for UnixFS directory sharding.
```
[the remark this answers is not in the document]
rvagg: _(I'll also note that https://github.com/aappleby/smhasher/issues/6 has the erroneous output for Guava as `4cae51b5316602c01c7c5642843e5fe7`, which doesn't figure in any of the above, it seems to be an actual error in that implementation)._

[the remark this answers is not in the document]
aschmahmann: > But is that something we want to document here?

@rvagg my understanding is that generally the that a major point of the multicodec table is to be able to have different groups reuse and implement the same schemes and identify their data the same way such that they can build compatible systems. In the case of multihash I'd generally assume that we're specifying data of the form `<hash code><hash len><hash bytes>` and how a user could take in a set of bytes and compute the same multihash output.

As a result, it seems like knowing how to order the output bytes to create the codec seems important for reproducibility. Am I missing something?

In general I'd hope that we could have some criteria for most codec entries which is to be able to point people at a spec (even if it's a living one) for how to implement these things and what the human meaning behind any particular 0xABCD actually means.

--- 

My main objections to what was merged here were:
1. Labeling 0x1022 as having a code despite it being both unspecified and unrequested
   - It doesn't personally matter to me what we specify, but I'd hope the next unfortunate soul to come along here and try to implement this could figure out whether the spaolacci approach (i.e. big endian) or the https://github.com/phensley/python-smhasher (i.e. small endian) was the one they should use to produce compatible results
   - We should be trying to make things better over time not worse, so let's try and push towards specifying what is meant by a new code.
2. Labeling 0x22 as murmur3-x64-64 despite knowing that this was incompletely specified and could still lead to confusion.
   - So let's label what we're doing. We can throw it in the notes field if we don't want to add more to the label name (and let any future soul who wants murmur3-x64-64-little-endian claim that if they want it).

PR #241, opened 2021-11-09
Allocate namespace for signalling ADLs for use in selectors/IPLD

willscott:
These are not used in CIDs regularly, so size is not a primary concern. As such, adding to a high byte range space.


warpfork:
So the idea is that these aren't used for codecs, but are magic numbers that may be used... somewhere else, in specific places where we're expecting to talk about ADLs, such as in a field of integer type in a selector extension we're about to propose?

willscott:
in general i would not expect them to be directly embedded as CIDs, but they are sufficiently cid-adjacent that they should not conflict with other defined multi-codecs, and I believe warrant their own namespace in the multicodec registry in which to be defined and delineated.

Stebalien (commented):
No reason not to merge these as draft codes.

Stebalien (approved):

Stebalien:
Nvm. Apparently we're not using codecs.

willscott:
this got deprecated in favor strings

PR #252, opened 2021-12-17
Add ISCC to multicodecs table

titusz:
The ISCC is similarity preserving identifier for digital content. An ISCC is derived algorithmically from the digital content itself, just like cryptographic hashes. However, instead of using a single cryptographic hash function to identify data only, the ISCC uses a variety of algorithms to create a composite identifier that exhibits similarity-preserving properties (softhash). For an implementation of code `0xcc01` see: https://core.iscc.codes/#quick-start

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -471,6 +471,7 @@ skein1024-1016,                 multihash,      0xb3df,         draft,
 skein1024-1024,                 multihash,      0xb3e0,         draft,
 poseidon-bls12_381-a2-fc1,      multihash,      0xb401,         permanent, Poseidon using BLS12-381 and arity of 2 with Filecoin parameters
 poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,     Poseidon using BLS12-381 and arity of 2 with Filecoin parameters - high-security variant
+iscc,                           softhash,       0xcc01,         draft,     ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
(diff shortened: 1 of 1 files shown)


rvagg:
Thanks @titusz, you'll have to be patient with us over the Christmas period with some of us taking time off, and this not being a straightforward entry to consider. Feel free to ping us again here if there's been no movement by the new year though.

titusz:
@vmx I was not sure what tag/category the ISCC would fit. I was tempted to use `multihash` as category. But there are two problems with that:

1. ISCC is not a cryptographic hash but a similarity preserving hash.
2. ISCC requires its own header format with varnibble encoded type, subtype, version, length fields.

So I went for a new category that a called `softhash`. It could also be called `finterprint`. You are right that `multiformat` might also fit as ISCC has its own typing system.

My idea was to make the ISCC multiformats compatible by wrapping its full byte digest with multibase, multicodec:

`<multibase><multicodec><iscc-digest-including-header>`

Any suggestions on how to approach this are welcome.

vmx:
> My idea was to make the ISCC multiformats compatible by wrapping its full byte digest with multibase, multicodec:
> 
> `<multibase><multicodec><iscc-digest-including-header>`

That sounds reasonable to me. Let's see what others say :)

sposth:
Hello, is there a timeline for deciding on this issue?

vmx (approved):
Sounds good to me, but I'd like to get one addition review from someone.

rvagg (approved):
Sorry, my fault on this being delayed so long, @vmx was pinging me to give a second opinion and I got a bit too busy. Thanks for your patience.
I think the discussion here is reasonable, this sounds like an experiment worth keeping an eye on and I don't see reason to object to this as it is.
There may be need to fiddle with the categorisation of this in the future, I see some reorg of what we're doing with the concept of grouping coming up at some point because it's not well defined or adhered to, but that shouldn't impact code or name which should stay stable as long as this continues to be useful for your project.

PR #263, opened 2022-04-06
Update table.csv

hangsiahong:
Add Kumandra to multicodecs table


rvagg (approved):
for anyone else looking, I found https://github.com/kumandra for this, Google wasn't much help for the name so I guess this is new & low-key

vmx (approved):

PR #274, opened 2022-06-24
feat: Add silverpine

Jorropo:
See https://github.com/Jorropo/go-silverpine

Right now there is no parameters to the `silverpine` component (all it need is an Ed25519 public key which it gets from a `p2p` component later).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -154,6 +154,7 @@ ed25519-priv,                   key,            0x1300,         draft,     Ed255
 secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
 x25519-priv,                    key,            0x1302,         draft,     Curve25519 private key
 kangarootwelve,                 multihash,      0x1d01,         draft,     KangarooTwelve is an extendable-output hash function based on Keccak-p
+silverpine,                     multiaddr,      0x3f42,         draft,     Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
 blake2b-8,                      multihash,      0xb201,         draft,     Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,         draft,


Jorropo:
Btw it seems we don't allow `è` in names.
Is that really a thing people care about ?
(if yes I'll just rename it to `silverpine`)

table.csv
```diff
@@ -152,6 +152,7 @@ sm2-pub,                        key,            0x1206,         draft,     SM2 p
 ed25519-priv,                   key,            0x1300,         draft,     Ed25519 private key
 secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
 x25519-priv,                    key,            0x1302,         draft,     Curve25519 private key
+pin-argenté,                    multiaddr,      0x3f42,         draft,     Experimental yggdrasil / ironwood routing protocol.
```
rvagg, line 155:
```suggestion
pin-argenté,                    multiaddr,      0x3f42,         draft,     Experimental yggdrasil / ironwood routing protocol
```

rvagg:
the name restriction:

> Codec names should be easily convertible to constants in common programming languages using basic transformation rules (e.g. upper-case, conversion of - to _, etc.). Therefore they should contain alphanumeric characters, with the first character being alphabetic. The primary delimiter for multi-part names should be -, with _ reserved for cases where a secondary delimiter is required. For example: bls12_381-g1-pub contains 3 parts: bls_381, g1 and pub, where bls_381 is "BLS 381" which is not commonly written as "BLS381" and therefore requires a secondary separator.

I'm not aware of a programming language (that's not ancient) that would have a problem with è, but characters outside of `[a-z0-9_-]` do tend to increase the friction for english-as-first-language developers (like me) who aren't used to entering those characters with their standard keyboards (I have to copy and paste è to get it in here, I have no idea how to do this with my keyboard). IDEs tend to make dealing with strange variable names easier so it's probably not as big a deal as it might have been once.

:shrug: if you feel strong enough about this I guess we could litigate this rule, pull in some others, and see if we can change it. I don't feel super strongly about it although I think I'd prefer the first character be in the `a-z` range (if I still have to find a è to copy and paste to get my IDE to auto-complete then that's as annoying as not having an IDE to auto-complete).

vmx:
> Btw it seems we don't allow `è` in names.
> Is that really a thing people care about ?

In some programming languages the table is is used to auto-generate some code. It's just simpler to use a restricted set of characters. I'd keep it that way to keep scripts simple.

Jorropo:
Renamed to `silverpine`.

rvagg (approved):

PR #282, opened 2022-08-23
Register IPFS and IPNS codecs

ben221199:
Previously, peers where described by `/ipfs/<peerId>`. This was replaced by `/p2p/<peerId>`, because that was a better name. Now, content is described by `/ipfs/<cid>` and `/ipns/<cid>`, but those were not defined in the Multicodec table.

This pull request will add `ipfs` and `ipns` to the Multicodec table and will remove the old `ipfs` used for describing peers. Hopefully, we also can get this definitions permanent.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -5,7 +5,7 @@ cidv2,                          cid,            0x02,           draft,     CIDv2
 cidv3,                          cid,            0x03,           draft,     CIDv3
 ip4,                            multiaddr,      0x04,           permanent,
 tcp,                            multiaddr,      0x06,           permanent,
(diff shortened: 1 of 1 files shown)


Stebalien:
These are already registered as "namespaces".

Stebalien:
See: ipfs-ns, ipld-ns, ipns-ns (0xe3, 0xe2, 0xe5).

ben221199:
There are raising multiple questions:
 - What is a namespace in the Multformat world? Where is it defined and specified?
 - Why the suffix `-ns`?
 - If you go with `ipfs-ns`, `ipns-ns` and `ipld-ns`, how to use it? Is it like this? `/ipfs-ns/<cid>`.

PR #283, opened 2022-08-24
rename namespaces and remove "ipfs" alias for "p2p"

Stebalien:
Previously, we called these foo-ns. Now we just call them foo (the fact that they're "namespaces" is implicit.

This way, the path component always matches the multicodec name (for both multiaddr and other path types).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -81,10 +81,10 @@ md4,                            multihash,      0xd4,           draft,
 md5,                            multihash,      0xd5,           draft,
 decred-block,                   ipld,           0xe0,           draft,     Decred Block
 decred-tx,                      ipld,           0xe1,           draft,     Decred Tx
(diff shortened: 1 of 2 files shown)


rvagg (approved):

ben221199:
Nice.

guseggert:
I would like to use human-readable codec names in an API but this makes me think twice, is it incorrect to consider these names as stable?

rvagg:
@guseggert mostly we try to keep them stable, but there are occasions like this when they get changed but I think in this case these names are not as likely to be used as they were anyway so the impact is considered fairly low. We're not about to go changing "dag-pb" or other commonly used names, however, even in cases where there's name-regret.

We also have this "draft" thing now, and these are drafts so that lessens the burden. Perhaps we don't document it but maybe there's an implicit "we won't touch this if it's 'permanent'".

PR #286, opened 2022-09-02
feat: beeson support

molekilla:
Adds BeeSon support

https://github.com/fairDataSociety/beeson


rvagg (approved):
I think this is fine, I was going to say this should be `serialization` but I note we have `cbor`, `json` and `raw` as `ipld` (I remember them being `serialization` but perhaps not..)

@vmx?

vmx (approved):
The categories are fuzzy. At least using `ipld` for this one seems to be consistent.

rvagg:
Ah, it actually has its own link type:

> The swarmCac and swarmSoc are misc types that are deserialized as regexed strings according to the rules of [Swarm CIDs](https://github.com/ethersphere/swarm-cid-js/). Additionally, the serialization can interpret the CID object used in swarm-cid-js.

presumably these could be interpreted as multiformats CIDs and therefore be viewed as actual IPLD; bonus.

molekilla:
thanks guys !

Issue #289, opened 2022-09-13
Register varsig codes


Gozala:
In https://github.com/ucan-wg/ucan-ipld/issues/4 we came to realize that we need self describing signatures for all the same reasons we want self describing hashes a.k.a [multihash](https://github.com/multiformats/multihash)es. 

Intention is to create repo with a `varsig` spec but in the meantime here is the intended format

```
<varint signature algorithm code><varint signature size in bytes><signature in bytes>
```

For that reason we would like to register:

1. codes for relevant signature algorithms
2. code for "varsig" umbrella signature code

Later will be used as secondary tag to enable custom not yet standardized signature algorithms and reduce rate of submissions here.

Gozala:
Also probably we can not use "multisignature" term as it would create confusion with multisigs. Maybe `autosig` would be a better term

rvagg:
sgtm

expede:
@Gozala with #290 merged, is there more to do in this Issue?

rvagg:
@Gozala got a plan for moving forward with this? Do you need a place for this, do you want a new repo or can we find a home in an existing repo. Multibase has an `rfcs` directory for "spec-like" things: https://github.com/multiformats/multibase/tree/master/rfcs, perhaps we could do something like that in one of the other multiformats repos?

Gozala:
> @Gozala got a plan for moving forward with this? Do you need a place for this, do you want a new repo or can we find a home in an existing repo. Multibase has an `rfcs` directory for "spec-like" things: https://github.com/multiformats/multibase/tree/master/rfcs, perhaps we could do something like that in one of the other multiformats repos?

This is already used by dag-ucan

https://github.com/ucan-wg/ucan-ipld/
https://github.com/ipld/js-dag-ucan

And there is an ongoing effort to spec out details here https://github.com/ChainAgnostic/varsig

Gozala:
Closing because PR with added codes had been merged

PR #290, opened 2022-09-14
Propose codes for signature algorithms

Gozala:
As per #289 this pr proposes to register common signature algorithms used in [UCAN](https://github.com/ucan-wg/spec/)s.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -486,9 +486,9 @@ zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcer
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
 fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
 varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
-EdDSA,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
-RS256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
-ES256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
+eddsa,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
+rs256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
+es256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
 plaintextv2,                    multiaddr,      0x706c61,       draft,
 holochain-adr-v0,               holochain,      0x807124,       draft,     Holochain v0 address    + 8 R-S (63 x Base-32)
 holochain-adr-v1,               holochain,      0x817124,       draft,     Holochain v1 address    + 8 R-S (63 x Base-32)
--- a/table.csv
+++ b/table.csv
@@ -96,7 +96,7 @@ bls12_381-g1g2-pub,             key,            0xee,           draft,     BLS12
 dash-block,                     ipld,           0xf0,           draft,     Dash Block
 dash-tx,                        ipld,           0xf1,           draft,     Dash Tx
 swarm-manifest,                 ipld,           0xfa,           draft,     Swarm Manifest
-swarm-feed,                     ipld,           0xfb,           draft,     Swarm 
+swarm-feed,                     ipld,           0xfb,           draft,     Swarm Feed
 beeson,                         ipld,           0xfc,           draft,     Swarm BeeSon
 udp,                            multiaddr,      0x0111,         draft,
 p2p-webrtc-star,                multiaddr,      0x0113,         draft,
--- a/table.csv
+++ b/table.csv
@@ -483,12 +483,12 @@ ssz,                            serialization,  0xb501,         draft,     Simpl
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,     SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
 iscc,                           softhash,       0xcc01,         draft,     ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
-fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
-fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
 varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
 eddsa,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
 rs256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
 es256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
+fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
+fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
 plaintextv2,                    multiaddr,      0x706c61,       draft,
 holochain-adr-v0,               holochain,      0x807124,       draft,     Holochain v0 address    + 8 R-S (63 x Base-32)
 holochain-adr-v1,               holochain,      0x817124,       draft,     Holochain v1 address    + 8 R-S (63 x Base-32)
--- a/table.csv
+++ b/table.csv
@@ -485,6 +485,10 @@ iscc,                           softhash,       0xcc01,         draft,     ISCC 
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
 fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
+varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
+EdDSA,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
+RS256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
+ES256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
 plaintextv2,                    multiaddr,      0x706c61,       draft,
 holochain-adr-v0,               holochain,      0x807124,       draft,     Holochain v0 address    + 8 R-S (63 x Base-32)
 holochain-adr-v1,               holochain,      0x817124,       draft,     Holochain v1 address    + 8 R-S (63 x Base-32)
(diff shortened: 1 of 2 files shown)


table.csv
```diff
@@ -485,6 +485,10 @@ iscc,                           softhash,       0xcc01,         draft,     ISCC
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent, Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
 fil-commitment-sealed,          filecoin,       0xf102,         permanent, Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
+varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
+EdDSA,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
+RS256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
+ES256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
```
Gozala, line 491:
```suggestion
varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
eddsa,                          varsig,         0xd001,         draft,     Edwards-Curve Digital Signature Algorithm
rs256,                          varsig,         0xd002,         draft,     RS256 Signature Algorithm
es256,                          varsig,         0xd003,         draft,     ES256 Signature Algorithm
```

table.csv
```diff
@@ -483,6 +483,12 @@ ssz,                            serialization,  0xb501,         draft,     Simpl
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,     SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
 iscc,                           softhash,       0xcc01,         draft,     ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
+varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
+es256k,                         varsig,         0xd0e7,         draft,     ES256K Siganture Algorithm (secp256k1)
+bls-12381-g1-sig,               varsig,         0xd0ea,         draft,     G1 signature for BLS-12381-G2
+bls-12381-g2-sig,               varsig,         0xd0eb,         draft,     G2 signature for BLS-12381-G1
+eddsa,                          varsig,         0xd0ed,         draft,     Edwards-Curve Digital Signature Algorithm
```
rvagg, line 490: why the gap of 1? are you anticipating something in between here?

table.csv
```diff
@@ -483,6 +483,12 @@ ssz,                            serialization,  0xb501,         draft,     Simpl
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,     SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
 iscc,                           softhash,       0xcc01,         draft,     ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
+varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
```
rvagg, line 486: I'm surprised you're not going to make a claim for a single-byte range for this, is that because these signatures are all going to be fairly large anyway so an extra byte isn't really a big deal?
Certainly makes this easier!

table.csv
```diff
@@ -496,3 +502,7 @@ skynet-ns,                      namespace,      0xb19910,       draft,     Skyne
 arweave-ns,                     namespace,      0xb29910,       draft,     Arweave Namespace
 subspace-ns,                    namespace,      0xb39910,       draft,     Subspace Network Namespace
 kumandra-ns,                    namespace,      0xb49910,       draft,     Kumandra Network Namespace
+es256,                          varsig,         0xd01200,       draft,     ES256 Signature Algorithm
```
rvagg, line 505: why are these all the way up here?

rvagg (approved):
lgtm; some of the numbering choices seem a little arbitrary, do you want to think a bit more about their placement and spacing before proceeding?

Gozala:
> lgtm; some of the numbering choices seem a little arbitrary, do you want to think a bit more about their placement and spacing before proceeding?

Thanks @rvagg! I think @expede had a reasoning to choose specific codes, but sadly I can't seem to find the thread with an elaboration. Personally I have ok with arbitrary codes as long as we have them & personally would prefer smaller ones.

@expede mind considering @rvagg's feedback here before we proceed ?

table.csv
expede: The format correlates to the equivalent public key in this table. We're "just" adding `0xd0` to the front of these ones.

| Description      | Type       | Code   |
| ---------------- | ---------- | ------ |
| bls12_381-g2-pub | Public Key | 0xeb   |
| bls12_381-g2-sig | Signature  | 0xd0eb |
| ed25519-pub      | Public Key | 0xed   |
| ed25519-sig      | Signature  | 0xd0ed |

table.csv
```diff
@@ -483,6 +483,12 @@ ssz,                            serialization,  0xb501,         draft,     Simpl
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,     SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
 iscc,                           softhash,       0xcc01,         draft,     ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,     0xcert Asset Imprint (root hash)
+varsig,                         varsig,         0xd000,         draft,     Namespace for all not yet standard signature algorithms
+es256k,                         varsig,         0xd0e7,         draft,     ES256K Siganture Algorithm (secp256k1)
+bls-12381-g1-sig,               varsig,         0xd0ea,         draft,     G1 signature for BLS-12381-G2
+bls-12381-g2-sig,               varsig,         0xd0eb,         draft,     G2 signature for BLS-12381-G1
+eddsa,                          varsig,         0xd0ed,         draft,     Edwards-Curve Digital Signature Algorithm
+eip-191,                        varsig,         0xd191,         draft,     EIP-191 Ethereum Signed Data Standard
```
expede, line 491: @rvagg This one is an exception to the above comment, since it uses secp256k1 but with a "nonstandard" (non-ECDSA) signature method.

table.csv
expede: There's a few places on this table that are like this, following the same pattern as we have here. The table enforces sorting by the code as if it were a number, not as sequence of bytes. `0xd999` is sorted as "lower" than `0xd0000`.

expede: These ended up separated because their public keys are also a byte longer than the other public key group 🤷‍♀️

expede: FWIW, I would love to get this table sorted as a byte array / lexicographically rather than numerically

table.csv
expede: A single byte range is probably too small if we want a consistent first digit. There are definitely more than 16 signature types out there. We're already at 10 in this PR, plus there's an additional 2 for Holochain that are not in this range (but are in the table).

table.csv
expede: Why no `0xd0ec`? Curve25519 is not currently widely used for signatures, with Ed25519 being preferred for this use case (you convert the Curve25519 key pair to Ed25519 and sign with EdDSA).

table.csv
expede: And indeed, we thought about compressing these into a tightly packed list. We definitely will need more than 16 (to accommodate signatures like EIP-191, possibly passkeys, and other nonstandard signature types).

I'm not completely against just counting up from `0xd000`, but the current layout is very easy to understand in relation to the existing PKs codes. As you say: relative to the size of a signature, we likely won't notice the extra byte in the 3-byte section.

table.csv
rvagg: > sorting by the code as if it were a number, not as sequence of bytes

Since the main consumption of values in this table is as integers encoded as varints, I think this is logical; I can see how you might want it otherwise but I don't think we have many uses that view these as bytes rather than numbers.

When someone eventually turns this into a .json, with the .csv a "compiled" form of it, we could potentially also compile other forms for different kinds of views of the same data where that's useful to folks.

table.csv
rvagg: Oh, so I misunderstood the intention of this one, I think. You're not really using it as a signal value to say "here follows a varsig", but a catch-all of some kind. Is the intent here that you're also making a claim on bytes starting with `0xd000` to `0xdfff` and we should be careful about allocating those in the future?

rvagg:
As per my comments above in the `varsig` entry, I'm not really clear on what the purpose of that entry is anymore, nor am I clear on why the byte forms are important for these. It doesn't seem to align with the proposal in #289 where we're doing varints, because the new comments here are about the importance of the bytes rather than the int codes. What am I missing?

expede:
> nor am I clear on why the byte forms are important for these [...] What am I missing?

@rvagg It makes them easier to remember for the humans using these codes. There's quite a lot of number punning in this list — for example, `0xec` for ECDSA and `0xed` for EdDSA.

> where we're doing varints

My understanding (as is the one in `did:key` AFAICT) is that the codes in this table can be encoded many ways, including but not limited to varints. The signature use case for a "varsig" would be encoded as a varint, but when looking up in this table we don't use varint, and a memotic is potentially helpful

> I'm surprised you're not going to make a claim for a single-byte range for this

Actually backing up — do you mean "why not register a single code in this table, and then have a separate standard for varint subcodes elsewhere?" If so, we had discussed that, but there appear to be ranges in this table for public keys, hashes, etc. But we could be wrong!

> You're not really using it as a signal value to say "here follows a varsig", but a catch-all of some kind.

That's a good point. @Gozala I think this was in your original proposal in `ucan-ipld` — thoughts?

> I'm not completely against just counting up from 0xd000,

@rvagg this really isn't an "over my dead body" -- if we're "holding it wrong", we can just update this PR to use `0xd000`, `0xd001`, `0xd002`, etc (or something). You tell me!

table.csv
Gozala: @expede I think you've misunderstood @rvagg comment, which if I'm understanding correctly simply asks why not pic a different, small code so it takes 1 byte as opposed to 3 used necessary for `0xdfff`.

@rvagg you are correct that this code is sort of catch-all and specifically it has been [defined as follows](https://github.com/ucan-wg/ucan-ipld/#251-nonstandard-signatures)

> ### 2.5.1 Nonstandard Signatures
>
> Nonstandard signature algorithms MAY be used, but are NOT RECOMMENDED as they are not widely interoperable.
>
> The `0xd000` prefix MUST be used if a nonstandard signature is used. Nonstandard signatures MUST be additionally prefixed with a varint of the signature length, and the UTF-8 UCAN `alg` field appended afterwards.
>
>```
> <varint 0xd000><varint sig_size><bytes sig><utf8 alg>`
> ```


I do not have strong preference between short and more memorable codes, but if I had to choose I'd probably side with former as these aren't meant to be memorized.

Gozala:
@expede @rvagg I have responded in https://github.com/multiformats/multicodec/pull/290#discussion_r981891960 but unless I"m misunderstanding / missing something we just need to decide between

1. Short codes
2. Codes that would map more closely with corresponding key codes

I personally think optimizing for size is probably better choice that optimizing for resemblance, that said I'm happy either way. But please let's move this forward as we're about to deploy code using this into production and it would be a lot more painful to change codes afterwards.

expede:
@rvagg @gozala to grease the wheels if we decide to switch to short codes, here's #293

rvagg:
Thanks for clarifying, I'm really just wanting to know that _you_ have clarity about how these are going to be used and whether it maps to these entries because it sounded a little like there might have been a disconnect. I'm fine with this PR as long as you both agree and can see the path forward from merging this PR. @Gozala has access to merge and this has been approved. If you want to go with #293 then that's fine too. No objections from me on either.

Gozala:
I've decided to merge this as 1 byte overhead seem negligible. If it becomes a problem we could always changes those codes as those are in draft status anyway

PR #291, opened 2022-09-15
Update with remianing signatures from ucan-ipld

expede:


expede:
@Gozala I guess that I don't have push rights onto branches on  `multiformats/multicodec`. Here's a branch into your PR from my fork.

table.csv
```diff
@@ -500,3 +502,7 @@ skynet-ns,                      namespace,      0xb19910,       draft,     Skyne
 arweave-ns,                     namespace,      0xb29910,       draft,     Arweave Namespace
 subspace-ns,                    namespace,      0xb39910,       draft,     Subspace Network Namespace
 kumandra-ns,                    namespace,      0xb49910,       draft,     Kumandra Network Namespace
+es256,                          varsig,         0xd01200,       draft,     ES256 Signature Algorithm
+es284,                          varsig,         0xd01201,       draft,     ES384 Signature Algorithm
+es512,                          varsig,         0xd01202,       draft,     ES512 Signature Algorithm
+rs256,                          varsig,         0xd01205,       draft,     RS256 Signature Algorithm
```
expede, line 508: Well that's annoying! `validate.py` takes the entire sequence as a number, so fails if we keep the signatures together 🤷‍♀️

PR #293, opened 2022-09-28
Sequential varsig codes

expede:


rvagg (approved):
here's a :thumbsup: if you and @Gozala decide to go with this first

PR #294, opened 2022-09-28
Sync multiaddr codecs

MarcoPolo:
These three codecs were defined in the multiaddr repo but not here. This syncs up the codec tables. I also added a checker to the multiaddr repo that will flag when this happens again in the future (https://github.com/multiformats/multiaddr/pull/142).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -25,6 +25,7 @@ murmur3-x64-64,                 multihash,      0x22,           permanent, The f
 murmur3-32,                     multihash,      0x23,           draft,
 ip6,                            multiaddr,      0x29,           permanent,
 ip6zone,                        multiaddr,      0x2a,           draft,
+ipcidr,                         multiaddr,      0x2b,           draft,     CIDR mask for IP addresses
 path,                           namespace,      0x2f,           permanent, Namespace for string paths. Corresponds to `/` in ASCII.
 multicodec,                     multiformat,    0x30,           draft,
 multihash,                      multiformat,    0x31,           draft,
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -130,6 +132,7 @@ messagepack,                    serialization,  0x0201,         draft,     Messa
 car,                            serialization,  0x0202,         draft,     Content Addressable aRchive (CAR)
 libp2p-peer-record,             libp2p,         0x0301,         permanent, libp2p peer record type
 libp2p-relay-rsvp,              libp2p,         0x0302,         permanent, libp2p relay reservation voucher
+memorytransport,                libp2p,         0x0309,         permanent, in memory transport for self-dialing and testing; arbitrar
```
mxinden, line 135:
```suggestion
memorytransport,                libp2p,         0x0309,         permanent, in memory transport for self-dialing and testing; arbitrary
```

See https://github.com/multiformats/multiaddr/blob/master/protocols.csv

mxinden (commented):
Thank you Marco!

table.csv
MarcoPolo: Oh no! I lost the last character in my copy paste. Thanks for finding him.

rvagg (approved):
just out of interest - why does `sni` need an entry?

MarcoPolo:
> just out of interest - why does `sni` need an entry?

We're using it in multiaddrs so that clients can set the SNI of a TLS handshake even if they've already resolved the domain name to an ip address.

See: https://github.com/multiformats/multiaddr/pull/138#issue-1365445793 and https://github.com/libp2p/go-libp2p/pull/1719#issue-1351286164. It's also part of the multiaddr README so the idea has been around for a while.

marten-seemann (approved):

PR #295, opened 2022-09-29
Add SRV multiaddr protocol

Pandapip1:
See https://github.com/multiformats/multiaddr/pull/143

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -98,7 +98,7 @@ dash-tx,                        ipld,           0xf1,           draft,     Dash 
 swarm-manifest,                 ipld,           0xfa,           draft,     Swarm Manifest
 swarm-feed,                     ipld,           0xfb,           draft,     Swarm Feed
 beeson,                         ipld,           0xfc,           draft,     Swarm BeeSon
-srv,                            multiaddr,      0xade,          draft,     DNS SRV Record (See RFC 2782, format: /srv/<servicename>/<tcp/udp/protocol>/<domain>, example: /srv/coolservice/tcp/domain.example.com)
+srv,                            multiaddr,      0xade,          draft,     "DNS SRV Record (See RFC 2782, format: /srv/<servicename>/<tcp/udp/protocol>/<domain>, example: /srv/coolservice/tcp/domain.example.com)"
 udp,                            multiaddr,      0x0111,         draft,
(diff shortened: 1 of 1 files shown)


rvagg:
@Pandapip1 see the validation failure here (you can run it locally), you'll need to fix up your comment to remove commas at least, try and keep it fairly simple and follow the patterns you can see in there for comments

marten-seemann (changes requested):
`/srv/coolservice/tcp/domain.example.com` seems to be the wrong way around. I have to parse the `tcp` argument before I know what to do with the `/src` component.
It doesn't work either, as `/tcp` is specified as a 16 bit field (to hold a port number), so you can't just put `domain.example.com` in there.

Pandapip1:
As you can tell, this is my first time. Thanks for the feedback!

One quick question: SRV is defined as a record: `_service._proto.domain`. `tcp` and `udp` are the most commonly used values for `proto`. How would I go about specifying that (for example) in `/srv/coolservice/tcp/domain.example.com`, the `/tcp/domain.example.com` is NOT an encapsulated multiaddr?

Pandapip1:
> try and keep it fairly simple and follow the patterns you can see in there for comments

Another quick question: where *do* you specify the format of new multiaddr protocols?

MarcoPolo:
> > try and keep it fairly simple and follow the patterns you can see in there for comments
> 
> Another quick question: where _do_ you specify the format of new multiaddr protocols?

I'd recommend reading https://github.com/multiformats/multiaddr/#interpreting-multiaddrs. Multiaddrs should be interpreted right to left. In this case I think the proper format of something like `_ldap._tcp.example.com` would be: `/dns/example.com/tcp/0/srv/ldap`. We interpret the multiaddr right to left, so we first see that this is a srv protocol, and the rest of the multiaddr (the stuff left of `srv`) makes sense in the context of `srv` (i.e. this is on example.com and a tcp addr) (note that `tcp/0` is chosen since we aren't specifying a specific tcp port, just that we want tcp). 

---

I'm curious, what's the use case here? I think outlining the use case helps everyone understand the motivation and get on the same page :)

Pandapip1:
> I'm curious, what's the use case here? I think outlining the use case helps everyone understand the motivation and get on the same page :)

If I'm being honest here, I actually forgot... I'll just close this until I re-remember it.

PR #298, opened 2022-10-28
add a new QUIC codepoint for QUIC v1

marten-seemann:
See https://github.com/multiformats/multiaddr/issues/145 for a detailed explanation.

Implementations MAY choose to continue accepting `/quic` multiaddresses (and interpret them as `quic-draft29` or `quic-v1` at their discretion).

---

cc @elenaf9 @thomaseizinger, who are apparently not part of the multiformats org yet.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -119,7 +119,7 @@ garlic32,                       multiaddr,      0x01bf,         draft,     I2P b
 tls,                            multiaddr,      0x01c0,         draft,
 sni,                            multiaddr,      0x01c1,         draft,     Server Name Indication RFC 6066 § 3
 noise,                          multiaddr,      0x01c6,         draft,
-quic-draft29,                   multiaddr,      0x01cc,         permanent,
+quic,                           multiaddr,      0x01cc,         permanent,
 quic-v1,                        multiaddr,      0x01cd,         permanent,
 webtransport,                   multiaddr,      0x01d1,         draft,
 certhash,                       multiaddr,      0x01d2,         draft,     TLS certificate's fingerprint as a multihash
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -119,7 +119,8 @@ garlic32,                       multiaddr,      0x01bf,         draft,     I2P b
 tls,                            multiaddr,      0x01c0,         draft,
 sni,                            multiaddr,      0x01c1,         draft,     Server Name Indication RFC 6066 § 3
 noise,                          multiaddr,      0x01c6,         draft,
-quic,                           multiaddr,      0x01cc,         permanent,
```
lidel, line 122: Removing `quic` name **will break all implementations that parse text multiaddrs based on libraries generated from this table**.  

@marten-seemann  is the plan to do a hard BREAKING CHANGE across ecosystem (forcing end users or implementations to update configs and DNS manually)  

OR 

do we want to have a deprecation period when both are valid, like we did when we renamed `/ipfs/{peerid}` to `/p2p/{peerid}`?   (fysa we've removed the legacy alias entry just recently in https://github.com/multiformats/multicodec/pull/283).

MarcoPolo: I don't think so. We should do the rename later and for now simply add the quic-v1 codepoint. (I left a similar comment in the multiaddr PR: https://github.com/multiformats/multiaddr/pull/147#discussion_r1016920915)

marten-seemann: I reverted this change.

thomaseizinger (approved):

elenaf9 (approved):

MarcoPolo (approved):

table.csv
guseggert: > We should do the rename later and for now simply add the quic-v1 codepoint.

I would advocate for not renaming things in this table anymore, ever. We have an increasing number of critical systems and businesses relying on these values remaining constant. There is enormous value in using the names as human-readable codes, but if we are unable to commit to not breaking names then I think we should just remove them entirely and rely on the description, which implies that it's for human and not machine consumption.

PR #302, opened 2022-11-17
Add EIP3770 chain-specific Ethereum address in table.csv 

qizhou:


rvagg:
How would this end up getting used in practice?

Let me pull in a few folks who I suspect may have some overlap between here and CAIP to figure out if this is the right approach; tbh I don't quite have the background to make much of a judgement on this. @oed, @ukstv, @clehner, @bumblefudge.

qizhou:
> How would this end up getting used in practice?
> 
> Let me pull in a few folks who I suspect may have some overlap between here and CAIP to figure out if this is the right approach; tbh I don't quite have the background to make much of a judgement on this. @oed, @ukstv, @clehner, @bumblefudge.

We plan to use it in web3:// access protocol (https://eips.ethereum.org/EIPS/eip-4804), where a name from ENS can be mapped to a chain-specific address of a contract that hosts a decentralized website.

ukstv:
@rvagg Thanks for mentioning!

@qizhou Why would you need multi codec to use with EIP-4804?

ukstv:
Maybe [CAIP-50](https://github.com/ChainAgnostic/CAIPs/pull/50/files) could work for your purposes?

qizhou:
> Maybe [CAIP-50](https://github.com/ChainAgnostic/CAIPs/pull/50/files) could work for your purposes?

Thanks!  Let me take a look.

oed:
I would love to see CAIP-50 pushed forward!

qizhou:
> TBH I'd prefer a binary namespace for the general pattern of VM:chainID:address, whether it's CAIP-50 or something else on the same pattern, rather than a separate one for each VM of the pattern chainID:address. in fact, half of CASA might throw tomatos at me for suggesting it, but even VM:chainID:publickey might be the more `multiformats` appropriate low-level expression, although then you get into the feature and/or bug of needing a signature from/connection to the wallet to derive said in EVM systems...
> 
> In any case, I think CAIP-50 has been languishing around looking for EXACTLY THIS use-case and a champion building on it, so @qizhou if you'd like to discuss this in a public CASA meeting let's get you calendared and invite the right interlocutors...

Sure.  Could you please send the invitation to me?  My email is qizhou@ethstorage.io.

BigLep:
2022-11-22 IPLD triage: check in on this in a future week or two to see how conversations progressed between interested parties. We may not need to make this specific change if it fits in CAIP-50.

rvagg:
@qizhou are we OK to close this out or have we not reached a satisfactory position with regard to the other CAIP work?

PR #303, opened 2022-11-26
Add CRC32 and CRC64 multihash

IllidanS4:


rvagg:
discussion about whether or not we can include non-cryptographic hashes is happening in #301; the same thing would apply here

vmx:
I don't know much about CRC, but after having a quick look the [CRC Wikipedia article](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) it seems that "CRC32" is well defined (ISO/IEEE etc.), but it's not so clear for CRC64. Would it make sense to include only CRC32 for now?

table.csv
```diff
@@ -108,6 +108,8 @@ p2p-circuit,                    multiaddr,      0x0122,         permanent,
 dag-json,                       ipld,           0x0129,         permanent, MerkleDAG json
 udt,                            multiaddr,      0x012d,         draft,
 utp,                            multiaddr,      0x012e,         draft,
+crc32,                          multihash,      0x0132,         draft,     CRC-32 non-cryptographic hash algorithm (IEEE 802.3)
+crc64,                          multihash,      0x0164,         draft,     CRC-64 non-cryptographic hash algorithm (ECMA-182 - Annex B)
```
rvagg, line 112:
```suggestion
crc32,                          hash,           0x0132,         draft,     CRC-32 non-cryptographic hash algorithm (IEEE 802.3)
crc64,                          hash,           0x0164,         draft,     CRC-64 non-cryptographic hash algorithm (ECMA-182 - Annex B)
```

as per https://github.com/multiformats/multihash/pull/156, if that's accepted

vmx (approved):
Thanks for the changes. I think I like `crc64-ecma` more.

rvagg (approved):

Issue #305, opened 2022-12-06
Add JSON form of the table


rvagg:
I can't find the issue(s) where we discussed this but there's been a proposal on the _table_ for a while that we represent the multicodec table as JSON for easier consumption and more flexibility. I'd like to switch it so that the CSV is generated from the JSON and people contribute to the JSON. A JSON table would let us add more items, like https://github.com/multiformats/multicodec/issues/297, and longer descriptions, and overall much more flexibility for entries and ease of downstream consumption.

This needs a PR to propose the format. Does it need to be line-delimited JSON? Can it be fully pretty-printed? What does linting look like? What does CSV generation look like?

Anyone want to give this a go?

zzo38:
I prefer to use CSV, and think it is simpler for many cases. (I do use the CSV myself.) However, it is also true that in some programming langauges, JSON will be better, so having both can be helpful.

One problem with using JSON for contribution is the lack of a trailing comma, meaning that the records do not all have the same format. Another note about JSON is that JSON does not properly have 64-bit integers; although JavaScript has a integer type now, JSON predates the integer type in JavaScript.

If JSON is implemented, ensure that the CSV is generated in the same format it is now (including spacing, although a column may be made wider if necessary), possibly adding additional columns at the end if such a thing is necessary (which it might or might not be). The CSV should be plain ASCII (do not use any non-ASCII characters), and should not have any quotation marks; if you convert JSON to CSV, ensure that this is the case, even though the JSON will definitely include quotation marks, and possibly non-ASCII characters. (If it includes long descriptions, they will need to somehow be truncated, or maybe it is better to have a separate field for a short description.)

If you do want to switch to JSON like that, you will first need to convert CSV to JSON first, so that it will all be JSON, and then you can convert JSON to CSV. (Then, you can check that the output matches the original, that the converter does not have a bug. Fortunately, you have a version control system, so you can revert it if something goes wrong.)

The program will have to check that it is valid. If the JSON includes the varint code (as suggested in #297), then it will have to check that the varint code matches the numeric code. This should be easily enough to implement.

[...]

PR #311, opened 2023-01-30
Add JSON form of table

rvagg:
Closes: #305

Draft for now, just putting it up for initial feedback.

* one entry per line, but still a valid complete JSON file, so it also has `[` and `]` lines.
* padded to make the columns line up
* `"description"` column is optional and is dropped where blank
* added `"intcode"` and `"varint"` (hex bytes) fields
* I also added an optional `"ref"` column with only the CID ones filled in so far .. I don't think left-padding this column is going to be ideal since `"description"` can get quite large. An option here is to add a compound object to combine `"description"` and `"ref"` and only left-pad the top-level items, `"status": "permanent", "notes": { "description": "CIDv1", "ref": [ "https://github.com/multiformats/cid" ] },`

What I think should be done to make this complete:

* [ ] Add a script that will compile JSON to CSV, correctly formatted
* [ ] Add a script that will load the CSV and write any updates back into the JSON - which would make for an easier workflow to add new entries where you don't want to have to deal with formatting or adding the `intcode` or `varint` entries
* [ ] Add a script that will lint both JSON and CSV and make sure they match, which we can also add to CI
* [ ] A Makefile that will run these scripts easily for any contributor and CI
* [ ] Docs to describe how to consume the data and how to contribute new entries


rvagg:
Link to file as you would view it on GitHub: https://github.com/multiformats/multicodec/blob/rvagg/json/table.json

rvagg:
Another alternative is to nest the 3 code values into a nested object: `"code": { "hex": "0x78", "int": 120, "varint": "0x78" }`. Any nested object doesn't get the padding treatment.

vmx:
Either layout is fine, I'd KISS and just have non-nested values. This also makes it easier if you e.g. write your own tooling and you want to convert it quickly to CSV.

PR #312, opened 2023-02-03
Add IPNS Record code

lidel:
This PR proposes adding `ipns-record`, like we already have one for `libp2p-peer-record`.

Rationale / potential uses:

-  Main one: we should add a dedicated codec for IPNS Record, otherwise people will use `ipns`(0xe5) code which would be invalid, as it indicates IPNS "namespace".
- [IPIP-351](https://github.com/ipfs/specs/pull/351)  will ship in Kubo 0.19 and will allow fetching signed (verifiable) IPNS Record.
  - In the future `/ipns/{id}?format=car` could return a CAR with `ipns-record` as an additional block.
- IPNI wants to support IPNS Records, but need codec (https://github.com/ipni/specs/pull/4)


lidel:
@rvagg correct. 
Block format will be this protobuf: https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-serialization-format

rvagg (approved):

PR #314, opened 2023-02-14
Adding multicodec for aes-gcm

guillaumemichel:
Adding `aes-gcm-256` multicodec. Please edit if the category `encryption` doesn't fit here.

`aes-gcm-256` is a payload format: the 12-byte Initialization Vector (IV) is concatenated with a payload encrypted using AES in Galois/Counter Mode, using a 256-bit key and the 12 byte IV.

This multicodec is required for the Double Hash DHT https://github.com/ipfs/specs/pull/373

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -69,6 +69,7 @@ aes-192,                        key,            0xa1,           draft,     192-b
 aes-256,                        key,            0xa2,           draft,     256-bit AES symmetric key
 chacha-128,                     key,            0xa3,           draft,     128-bit ChaCha symmetric key
 chacha-256,                     key,            0xa4,           draft,     256-bit ChaCha symmetric key
+aes-gcm-256,                    encryption,     0xa5,           draft,     AES Galois/Counter Mode with 256-bit key and 12-byte IV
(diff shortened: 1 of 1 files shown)


vmx (commented):
Sounds reasonable to me. I've had a brief look at the corresponding section in the Double Hash DHT spec. Please note that the encoded value won't be `0xa5`, but the varint of it: `0xa501`.

Though I'd like to get input from others as this introduces a new category.

rvagg (changes requested):
Since this opens the door to a massive number of permutations, all under the same "encryption" category, I think this should go later in the table where there is a sizeable chunk of free space.

Otherwise I don't mind a new category .. there's just a lot of them and I can imagine the inevitable PR with someone adding _all the possible permutations_ ..

guillaumemichel:
Moved `aes-gcm-256` to `0x2000`

```
code               varint
8192 (0x2000)   => 10000000 01000000 (0x8040)
```

rvagg (approved):
nice 👌

PR #315, opened 2023-02-16
adds multicodec table entries for private keys of ecdsa curves

kezike:
This PR adds multicodec table entries for private keys of ecdsa curves P-256, P-384, and P-521.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -163,9 +163,9 @@ ed25519-priv,                   key,            0x1300,         draft,     Ed255
 secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
 x25519-priv,                    key,            0x1302,         draft,     Curve25519 private key
 rsa-priv,                       key,            0x1305,         draft,     RSA private key
-p256-priv,                      key,            0x1306,         draft,     P-256 private key (compressed)
-p384-priv,                      key,            0x1307,         draft,     P-384 private key (compressed)
-p521-priv,                      key,            0x1308,         draft,     P-521 private key (compressed)
+p256-priv,                      key,            0x1306,         draft,     P-256 private key
+p384-priv,                      key,            0x1307,         draft,     P-384 private key
+p521-priv,                      key,            0x1308,         draft,     P-521 private key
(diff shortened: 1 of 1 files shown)


rvagg:
this is fine I think, but what does it mean for the private keys to be "compressed"? is there a compressed form of them and an uncompressed form that we're differentiating in the private keys?

vmx (approved):
The compression is about different representations of the same data. I don't see much reason to use the uncompressed form of keys anywhere.

conr2d:
AFAIK a private key on elliptic curve is a scalar value, and there is no compressed form unlike its corresponding public key.

kezike:
Thanks for the feedback all. I only added `compressed` to follow the example of the other ECDSA entries (`p256-pub`, `p384-pub`, and `p521-pub`). I wasn't sure whether I should use it for the private keys as well. If not, I can remove it.

table.csv
```diff
@@ -163,6 +163,9 @@ ed25519-priv,                   key,            0x1300,         draft,     Ed255
 secp256k1-priv,                 key,            0x1301,         draft,     Secp256k1 private key
 x25519-priv,                    key,            0x1302,         draft,     Curve25519 private key
 rsa-priv,                       key,            0x1305,         draft,     RSA private key
+p256-priv,                      key,            0x1306,         draft,     P-256 private key (compressed)
+p384-priv,                      key,            0x1307,         draft,     P-384 private key (compressed)
+p521-priv,                      key,            0x1308,         draft,     P-521 private key (compressed)
```
rvagg, line 168:
```suggestion
p256-priv,                      key,            0x1306,         draft,     P-256 private key
p384-priv,                      key,            0x1307,         draft,     P-384 private key
p521-priv,                      key,            0x1308,         draft,     P-521 private key
```

I don't believe there's a distinction here for the private key form so best to remove this unless someone wants to point me to a variation where a private key can either have a compressed or uncompressed form. Other than this I'm fine with these in here.

rvagg (changes requested):

table.csv
kezike: Done ✅

rvagg (approved):

PR #321, opened 2023-03-30
Add a transport code to indicate content is provided via car

willscott:
per the trustless gateway spec

The exact semantics of this spec are evolving, but might as well reserve a code that we can use for indicating the existence of an HTTP endpoint for index providers.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -143,7 +143,7 @@ car-index-sorted,               serialization,  0x0400,         draft,     CARv2
 car-multihash-index-sorted,     serialization,  0x0401,         draft,     CARv2 MultihashIndexSorted index format
 transport-bitswap,              transport,      0x0900,         draft,     Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,     Filecoin graphsync datatransfer
-transport-gateway-trustless,    transport,      0x0920,         draft,     HTTP car datatransfer
+transport-ipfs-gateway-http,    transport,      0x0920,         draft,     HTTP car datatransfer
 multidid,                       multiformat,    0x0d1d,         draft,     Compact encoding for Decentralized Identifers
 sha2-256-trunc254-padded,       multihash,      0x1012,         permanent, SHA2-256 with the two most significant bits from the last byte zeroed (as via a mask with 0b00111111) - used for proving trees as in Filecoin
 sha2-224,                       multihash,      0x1013,         permanent, aka SHA-224; as specified by FIPS 180-4.
--- a/table.csv
+++ b/table.csv
@@ -143,7 +143,7 @@ car-index-sorted,               serialization,  0x0400,         draft,      CARv
 car-multihash-index-sorted,     serialization,  0x0401,         draft,      CARv2 MultihashIndexSorted index format
 transport-bitswap,              transport,      0x0900,         draft,      Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,      Filecoin graphsync datatransfer
-transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP car datatransfer
+transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP IPFS Gateway trustless datatransfer
 multidid,                       multiformat,    0x0d1d,         draft,      Compact encoding for Decentralized Identifers
 sha2-256-trunc254-padded,       multihash,      0x1012,         permanent,  SHA2-256 with the two most significant bits from the last byte zeroed (as via a mask with 0b00111111) - used for proving trees as in Filecoin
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -143,6 +143,7 @@ car-index-sorted,               serialization,  0x0400,         draft,     CARv2
 car-multihash-index-sorted,     serialization,  0x0401,         draft,     CARv2 MultihashIndexSorted index format
 transport-bitswap,              transport,      0x0900,         draft,     Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,     Filecoin graphsync datatransfer
+transport-gateway-trustless,    transport,      0x0920,         draft,     HTTP car datatransfer
```
masih, line 146: With "gateway" in the name, am i right to assume this is for car transfer over IPFS HTTP gateway? 
If so, would it make sense to include `ipfs` in the name?

```suggestion
transport-ipfs-gateway-trustless,    transport,      0x0920,         draft,     HTTP IPFS Gateway CAR datatransfer
```

rvagg: yes, if we view this namespace as having a very broad audience, making the assumption that "gateway" bears the singular meaning we think it does isn't appropriate
+1

rvagg:
how will this code end up getting used in the network chatter?

table.csv
aschmahmann: Why would you restrict this to CAR?

willscott:
> how will this code end up getting used in the network chatter?

my expectation is that this will be used in the same context as `transport-bitswap` and `transport-filecoin-graphsync-v1` - it will be a code understood in context routing contexts to indicate that this publisher can provide data over http.

rvagg:
The only remaining hold-up here is the use of "gateway" unadorned, since that has a specific meaning for _us_ but not all audiences of this table, `ipfs-gateway` would seem to be the appropriate way to solve that. Other than that it's good to go I think. I don't know about the comments about "CAR" but since that's in the description, it can be easily changed later if necessary.

hannahhoward:
I'd support adding IPFS -- @willscott ?

willscott:
updated name to be less ambiguous

masih (approved):

table.csv
aschmahmann: > do you mean versus also including single blocks, or versus also including unverifiable responses?

I meant single blocks

aschmahmann: Note: I would prefer knowing what this thing actually is before we merge it as it seems like there are a bunch of unknowns as to what this actually means/how it will be used and the point of the shared table is to enable interoperability (if it wasn't then people could just have system-specific enums and call it a day).

If the desire of the crowd here is to merge this PR before there are any definitions I'd like to request that the entry be removed from the table should it arise that this entry turns out to be abandoned or there is no definition of it in X (maybe 3 or as high as 6?) months.

No we haven't held other entries to this bar (especially the really old ones), but it'd be nice if we could do better going forward here and it doesn't seem like a huge ask that more than the people involved in this PR would be able to figure out what is meant by a the code being added to the table.

willscott: We can change the comment to
`HTTP IPFS Gateway trustless datatransfer`
is there other wording at this stage you would propose / prefer?

I'm happy to re-visit in 3-6 months to
(a) remove if we aren't using it
(b) revising wording to reflect / point to the specification we've ended up with

rvagg (approved):

table.csv
```diff
@@ -143,6 +143,7 @@ car-index-sorted,               serialization,  0x0400,         draft,      CARv
 car-multihash-index-sorted,     serialization,  0x0401,         draft,      CARv2 MultihashIndexSorted index format
 transport-bitswap,              transport,      0x0900,         draft,      Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,      Filecoin graphsync datatransfer
+transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP IPFS Gateway trustless datatransfer
```
lidel, line 146: :thought_balloon:  This terribly long string will be sent with every IPNI roundtrip, and due to the way IPNI implemented /routing/v1 API, we use string that can change at atny time, and not the number: https://github.com/ipfs/specs/issues/377

I'd be strongly recommending moving to number instead of string, which we did change multiple times in the past ( `/ipfs/` → `/p2p`, then `/ipns-ns` → `/ipns` etc)

willscott: ℹ️ the cid.contact find response https://github.com/ipni/specs/blob/main/schemas/v1/openapi.yaml#L120 does use compact number versions

masih: IPNI [metadata](https://github.com/ipni/specs/blob/main/IPNI.md#metadata) always uses the varint numerical value.

lidel: sgtm, I've opened https://github.com/ipfs/specs/pull/403 for explicitly supporting hex codes on `/routing/v1`

PR #322, opened 2023-04-06
Add Varsig Multiformat

expede:
Hey all 👋 We ended up wanting a LOT more flexibilty in Varsig than a codec tag, and broke it out into its own spec here: https://github.com/ChainAgnostic/varsig

This is essentially `multisig`, except that sequence of letters has a widely-used namespace conflict in multisig wallets. To avoid confusion, it's "varsig", but almost certainly belongs in the block with the other multiformats.

If useful, we can remove the previously-added `draft` signature types from this table, but I know  historically there's been resistence to that. I'm happy either way 👍

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -506,7 +506,7 @@ ssz-sha2-256-bmt,               multihash,      0xb502,         draft,      SSZ 
 json-jcs,                       ipld,           0xb601,         draft,      The result of canonicalizing an input according to JCS - JSON Canonicalisation Scheme (RFC 8785)
 iscc,                           softhash,       0xcc01,         draft,      ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,      0xcert Asset Imprint (root hash)
-varsig,                         varsig,         0xd000,         draft,      Namespace for all not yet standard signature algorithms
+nonstandard-sig,                varsig,         0xd000,         draft,      Namespace for all not yet standard signature algorithms
 es256k,                         varsig,         0xd0e7,         draft,      ES256K Siganture Algorithm (secp256k1)
 bls-12381-g1-sig,               varsig,         0xd0ea,         draft,      G1 signature for BLS-12381-G2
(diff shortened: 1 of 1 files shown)


vmx (commented):
I'm not deep enough in the signature business, though it sounds good to me. Though adding a 1 byte codec certainly needs some thought, I think adding a new "multi*" is good use of it.

expede:
Awesome thanks `@vmx` :) @rvagg any objections?

rvagg:
sgtm; although "nonstandard-sig" - is this just a legacy thing now and the rename is a way of deprecating it? If so, should we just add a note in the description column that it's going to be removed, or at least that it's not intended to be used for anything going forward?

expede:
> should we just add a note in the description column that it's going to be removed

Yeah totally — happy to deprecate those. Will do shortly

table.csv
```diff
@@ -505,7 +506,7 @@ ssz-sha2-256-bmt,               multihash,      0xb502,         draft,      SSZ
 json-jcs,                       ipld,           0xb601,         draft,      The result of canonicalizing an input according to JCS - JSON Canonicalisation Scheme (RFC 8785)
 iscc,                           softhash,       0xcc01,         draft,      ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,      0xcert Asset Imprint (root hash)
-varsig,                         varsig,         0xd000,         draft,      Namespace for all not yet standard signature algorithms
+nonstandard-sig,                varsig,         0xd000,         draft,      Namespace for all not yet standard signature algorithms
```
rvagg, line 509:
```suggestion
nonstandard-sig,                varsig,         0xd000,    deprecated,      Namespace for all not yet standard signature algorithms
```

@vmx new status column type, indicating eventual removal?

vmx: Works for me.

expede: Works for me too!

rvagg (approved):

PR #324, opened 2023-05-24
Add httpath multiaddr component

rvagg:
Ref: https://github.com/ipni/go-libipni/blob/dfad9b30814eaede8c7e934f120b632bb180511c/maurl/convert.go#L35
Ref: https://github.com/ipni/go-libipni/issues/42


willscott (approved):

lidel (changes requested):
For the record, me and @MarcoPolo raised concerns in https://github.com/ipni/go-libipni/issues/42, would be good to keep this open until we understand there is utility of this outside of IPNI and using `https://` URLs.

MarcoPolo:
Let's do http-path https://github.com/multiformats/multiaddr/pull/164 instead

PR #325, opened 2023-06-02
Add code for SCION multiaddrs

leonrinkel:
We'd like to upstream libp2p support for the [SCION](https://scion-architecture.net) Internet architecture. As a first step, this PR reserves a code for SCION multiaddrs like `/scion/19-ffaa:1:1079...`.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -531,3 +531,4 @@ es256,                          varsig,         0xd01200,       draft,      ES25
 es284,                          varsig,         0xd01201,       draft,      ES384 Signature Algorithm
 es512,                          varsig,         0xd01202,       draft,      ES512 Signature Algorithm
 rs256,                          varsig,         0xd01205,       draft,      RS256 Signature Algorithm
+scion,                          multiaddr,      0xd02000,       draft,      SCION Internet architecture


mxinden (commented):
No objections from my end.  Interesting project.

That said, what is the benefit for tracking it here without an implementation? Why not experiment with the [Private Use Range](https://github.com/multiformats/multicodec#private-use-area) first and only reserve the multicodec once there is a first implementation using it?

rvagg:
I'll wait for @leonrinkel's response to the above question, but wrt multiaddr and private range, we've recently had an issue with IPNI using a private range for `httpath`, in the wild, for some time now (although it seems like we might be heading in a different direction with this https://github.com/multiformats/multicodec/pull/324). Since multiaddrs get encoded in binary form, it does matter that the number is encoded in the table.
So I think it's fine to go ahead with this as long as it seems reasonable that this isn't vapourware and we'll be stuck with an entry that never gets used at all.

leonrinkel:
We do have an implementation that we're experimenting with and I'm trying to upstream it step by step. I just thought it would be the right order to do this first. I'll open a multiaddr PR asap, it's only minor version bumps in the SCION net library that I'm still waiting for. But I see your point, if you want to wait till then and cross-reference, that is also fine for me.

rvagg (approved):
since there's not an explicit rejection from the libp2p team I see no reason to hold this up. @leonrinkel if this ends up going nowhere then it'd be great if you could come back and PR for a removal, but it seems to me from what you've said and the existence of code that this has a future in an attempt at least, which is good enough.

PR #326, opened 2023-06-05
Fix typo: eth-reciept-log → eth-receipt-log

darrylyeo:


rvagg (approved):
@i-norden can you sanity-check here, are you using these names in a way that would break if we change them here?

i-norden:
This LGTM, won't break anything on our end and we use the correct name [elsewhere](https://github.com/vulcanize/ts-dag-eth/blob/main/log/src/index.ts#L7) so will be good to have this match. Looks like I was who introduced this typo 😅 thanks for the fix!

PR #328, opened 2023-06-09
Rename urdca-2015-canon to rdfc-1-0

peacekeeper:
The [W3C RDF Dataset Canonicalization and Hash Working Group](https://www.w3.org/groups/wg/rch) has decided that the name of the algorithm should be RDFC-1.0. We therefore request that this be updated in this table.

See these links for more context:

- https://www.w3.org/2023/06/07-rch-minutes.html#r01
- https://github.com/w3c/rdf-canon/issues/88
- https://github.com/w3c/rdf-canon/pull/116

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -504,7 +504,7 @@ xxh3-64,                        hash,           0xb3e3,         draft,      Extr
 xxh3-128,                       hash,           0xb3e4,         draft,      Extremely fast non-cryptographic hash algorithm
 poseidon-bls12_381-a2-fc1,      multihash,      0xb401,         permanent,  Poseidon using BLS12-381 and arity of 2 with Filecoin parameters
 poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,      Poseidon using BLS12-381 and arity of 2 with Filecoin parameters - high-security variant
-urdca-2015-canon,               ipld,           0xb403,         draft,      The result of canonicalizing an input according to URDCA-2015 and then expressing its hash value as a multihash value.
+rdfc-1-0,                       ipld,           0xb403,         draft,      The result of canonicalizing an input according to RDFC-1.0 and then expressing its hash value as a multihash value.
 ssz,                            serialization,  0xb501,         draft,      SimpleSerialize (SSZ) serialization
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,      SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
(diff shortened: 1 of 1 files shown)


vmx (approved):

rvagg:
Seems fine to me, but we should ping folks from https://github.com/multiformats/multicodec/pull/261 to make sure we're not stepping on toes too hard with the straight name change here: @dmitrizagidulin, @dlongley, @msporny, @gobengo, @RangerMauve

table.csv
```diff
@@ -501,7 +501,7 @@ xxh3-64,                        hash,           0xb3e3,         draft,      Extr
 xxh3-128,                       hash,           0xb3e4,         draft,      Extremely fast non-cryptographic hash algorithm
 poseidon-bls12_381-a2-fc1,      multihash,      0xb401,         permanent,  Poseidon using BLS12-381 and arity of 2 with Filecoin parameters
 poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,      Poseidon using BLS12-381 and arity of 2 with Filecoin parameters - high-security variant
-urdca-2015-canon,               ipld,           0xb403,         draft,      The result of canonicalizing an input according to URDCA-2015 and then expressing its hash value as a multihash value.
+rdfc-1-0,                       ipld,           0xb403,         draft,      The result of canonicalizing an input according to RDFC-1.0 and then expressing its hash value as a multihash value.
```
msporny, line 504:
```suggestion
rdfc-1,                       ipld,           0xb403,         draft,      The result of canonicalizing an input according to RDFC-1.0 and then expressing its hash value as a multihash value.
```

I expect that rdfc-1 is just fine, and a point-release would be backwards-compatible. If we need to, we can do rdfc-1.1/rdfc-1-1 in the future. Suggest just keeping it to a major number for now.

dlongley: I think `rdfc-1` is fine here. Any changes to the algorithm wouldn't change outputs (without changing a major version) but rather just be some kind of spec fixes for minor versions, so we only need one codec per major version, IMO.

dmitrizagidulin (approved):
👍 from me, thanks.

table.csv
dmitrizagidulin: Agreed re `rdfc-1`

dlongley: @peacekeeper, if you're able to apply this suggestion to your PR I think it's good to merge.

BigLep:
@peacekeeper: are you able to incorporate PR feedback so this can be merged?

peacekeeper:
Oops sorry this took so long. I incorporated the feedback, should be ready now I think.

dlongley (approved):

vmx:
A majority for people involved is happy. I had to rebase it, but the change is identical. I'm merging this now. Thanks everyone for diving into this and the previous issue around this.

PR #329, opened 2023-06-13
Add sha256a

oed:
This PR adds sha256a which is an associative hash function. It's defined as the sum of multiple sha2-256 hashes.
You can read the [specification here](https://cips.ceramic.network/CIPs/cip-124#sumofsha256s).

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -175,6 +175,7 @@ kangarootwelve,                 multihash,      0x1d01,         draft,      Kang
 aes-gcm-256,                    encryption,     0x2000,         draft,      AES Galois/Counter Mode with 256-bit key and 12-byte IV
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
+sha256a,                        multihash,      0x7012,         draft,      The sum of multiple sha2-256 hashes
 blake2b-8,                      multihash,      0xb201,         draft,      Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,         draft,
 blake2b-24,                     multihash,      0xb203,         draft,
--- a/table.csv
+++ b/table.csv
@@ -175,7 +175,7 @@ kangarootwelve,                 multihash,      0x1d01,         draft,      Kang
 aes-gcm-256,                    encryption,     0x2000,         draft,      AES Galois/Counter Mode with 256-bit key and 12-byte IV
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
-sha256a,                        multihash,      0x7012,         draft,      The sum of multiple sha2-256 hashes; as specified by Ceramic CIP-124.
+sha256a,                        hash,           0x7012,         draft,      The sum of multiple sha2-256 hashes; as specified by Ceramic CIP-124.
 blake2b-8,                      multihash,      0xb201,         draft,      Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,         draft,
 blake2b-24,                     multihash,      0xb203,         draft,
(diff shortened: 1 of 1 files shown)


table.csv
```diff
@@ -175,6 +175,7 @@ kangarootwelve,                 multihash,      0x1d01,         draft,      Kang
 aes-gcm-256,                    encryption,     0x2000,         draft,      AES Galois/Counter Mode with 256-bit key and 12-byte IV
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
+sha256a,                        multihash,      0x7012,         draft,      The sum of multiple sha2-256 hashes
```
vmx, line 178:
```suggestion
sha256a,                        multihash,      0x7012,         draft,      The sum of multiple sha2-256 hashes; as specified by Ceramic CIP-124.
```
I know the "Ceramic" is redundant with the "C" in "CIP", but my hope would that this way it's easy for people to find the spec.

oed: Thanks, wasn't sure about convention about adding information on where to find a spec.

vmx: There isn't really a convention. I just looked what we did in other cases :)

vmx (approved):
It looks good to me, but I'd like to also have an approval from @rvagg before merging.

oed:
Ping @rvagg :)

rvagg (approved):
Wild, I think we're stretching the definition of a "hash function" with this one if we use a classic `fn(content)=digest` definition. What's the "content" in this case? If you used this in a CID, what would that mean? It's got some pretty big downsides if used in place of a standard cryptographic hash function - if you broke up content, did a sha256 across each chunk, then combined them with sha256a then you'd have a "digest" that's valid for all possible ordered recombinations of those chunks.

I'm finding it hard to say no, but this is not a comfortable fit I think.

oed:
> What's the "content" in this case?

All of the data the sum of which is this hash.

> If you used this in a CID, what would that mean?

Not sure, this is not the intention. Goal here is to use this as a way to have upgradability if we need to use another ahash function (using something else than sha256 under the hood)

> if you broke up content, did a sha256 across each chunk, then combined them with sha256a then you'd have a "digest" that's valid for all possible ordered recombinations of those chunks.

This is actually the exact reason why we are using it (see our [Recon](https://cips.ceramic.network/CIPs/cip-124) spec, or the "Range-Based Set Reconciliation and Authenticated Set Representations” [[arXiv:2212.13567](https://doi.org/10.48550/arXiv.2212.13567)] paper where the idea comes from)

aschmahmann:
> What's the "content" in this case?
>> All of the data the sum of which is this hash.

IIUC the distinction here is that most hash functions (i.e. all of the ones currently covered by multihash) take in bytes and spit out a hash. However, it seems like sha256a takes a list of bytes and spits out a hash which is a different abstraction.

While you could extract the part the part of sha256a that does bytes -> hash mapping, that would be the part that divides up a byte array into 32 bytes chunks and sums them up, which doesn't have much to do with sha256 which is largely what makes this awkward.

aschmahmann:
@rvagg @vmx is the guidance in use for multihash still https://github.com/multiformats/multihash#non-cryptographic-hash-functions? Since collisions are an intentional part of the design of `sha256a` it seems like `hash` would be the tag to use (leaving aside the other issue of this not quite fitting the bytes -> digest abstraction hashes normally use).

rvagg:
@aschmahmann yeah, it's a bit ambiguous in that it builds on a cryptographic hash function but then undoes some of the guarantees provided! The text in your link that's probably most relevant is:

> are not suitable for content addressing systems

By that definition, if this isn't intended for use as a CID for addressing (acknowledging that there's even ambiguity in how CIDs get used in the wild, so even that isn't a clear demarcation!), then switching it to `hash` might be a better choice here to (a) avoid confusion and side-step ambiguity and (b) not set a problematic precedent, which is a common problem we have with the table.

Would you mind switching to the `hash` tag @oed? I think we'd be good with that.

oed:
Makes sense, updated to `hash`. Thanks @rvagg @aschmahmann

PR #330, opened 2023-06-17
feat: allocating inline IPLD multiformity

mikeal:
We've talked about this informally for a while, I'd like to start experimenting with it but I want to make sure we get low numbers allocated to basic inline IPLD types.


mikeal:
We may also want to allocate uint8, uint16, and uint32 multiformats for when we need to do fixed length number encodings.

rvagg:
bit of a can of worms here if we start adding all the possible types in the single byte range - various string encodings, float encodings, all the ints and their endianness forms; so maybe this should be done in the two-byte range so we can add whatever people want?

PR #332, opened 2023-07-29
Add LBRY protocol

ben221199:
This pull request adds a multiaddr for the LBRY network protocol, as described at https://lbry.tech/spec.


table.csv
```diff
@@ -54,6 +54,7 @@ leofcoin-pr,                    ipld,           0x83,           draft,      Leof
 sctp,                           multiaddr,      0x84,           draft,
 dag-jose,                       ipld,           0x85,           draft,      MerkleDAG JOSE
 dag-cose,                       ipld,           0x86,           draft,      MerkleDAG COSE
+lbry,                           multiaddr,      0x8c,           draft,      LBRY Address
```
Stebalien, line 57: This should likely be in the `namespace` category, not the `multiaddr` category. Like swarm, ipfs, ipld, zeronet, dnslink, etc.

ben221199: I see that `ipfs` and `ipns` also use `namespace` type indeed, so I should change it for that consistancy, but what is actually the difference between the two? I mean, both `multiaddr`s like `ip4` and `namespace`s like `ipfs` are some sort of address that is used in the `/<name>/<optionalValue>` form? Is there any Multiformats documentation that describes the different types?

Stebalien: See https://github.com/multiformats/multiaddr/pull/161#issuecomment-1657199288.

Basically, multiaddr is specifically for addressing "endpoints" and it's designed to be recursive. I.e., speak protocol X over protocol Y over protocol Z.

Stebalien: Basically, a multiaddr is a set of instructions for _how_ to connect to some endpoint (which is why we use it in libp2p). Here it looks like you want a namespace for naming content.

ben221199: So actually, I mistook the `/<name>/<optionalValue>` format for being `multiaddr` specific, and that format is actually `namespace` specific, but because `multiaddr` is a `namespace`, `namespace` and `multihash` are a `multicodec`, and `multicodec` and `multibase` are `multiformats`, it actually makes sense that `multiaddr` uses the same format?

image

I think as Multiformats community we should really think through about the types, registries and the hierarchy of all those types and registries, because even for me it becomes to start confusing. Only for `multibase` it is clear at the moment.

I wanted to make an issue on https://github.com/multiformats/multiformats, but I see I already did that a while ago: https://github.com/multiformats/multiformats/issues/63. We can continue the discussion there.

For now, I just will change the `tag` of `lbry` to `namespace` to follow the others, because I simply want this thing registered.

Stebalien: Oh, I totally agree. I actually want to reorganize it all into a single "multipath" spec where all of these are defined as `<varint-encoded-multicodec><arbitrary-data>` (that's effectively what we have anyways).

Really, "namespace" is just the most general usage of multicodecs. It's the "everything else" category for when you need a way to prefix some form of "name" but it doesn't fit into any of the other categories.

ben221199:
Changed tag from `multiaddr` to `namespace`. I think it is fine now.

table.csv
ben221199: Yeah, lets see what the future brings us.

ben221199:
Can it be merged, @rvagg?

vmx (approved):
As @Stebalien already reviewed this and I had a quick look I dare to approve and merge it.

ben221199:
Thanks @vmx. I would be happy to see it merged. 😃

Issue #342, opened 2024-02-15
Proposal: s3-sha2-256-x for Amazon's parallel checksums


drernie:
We would like to use a multihash for the new [Amazon S3 Checkums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/), specifically their SHA-256 variant.
Is this the right place and way to request a new prefix for that?

These differ from traditional sha-256 in that they are computed in parallel using a variable block size.
Being amazon, some of their SDKs use 8mb and others use 5mb (though it is user-configurable).

Specifically, I would like to propose (by analogy with `dbl-sha2-256` and `ssz-sha2-256-bmt`):

```
s3-sha2-256-5,                   multihash,      0xb515,           draft,
s3-sha2-256-8,                   multihash,      0xb518,           draft,
```

This could be used both for the native S3 algorithm, as well as compatible implementations (such as ours).

I believe this satisfies the "two implementations" requirement for [mulitformats registration](https://github.com/multiformats/multiformats/blob/master/contributing.md#multiformats-registrations).
Is that sufficient? Should I just create a PR, or is there a more formal process?

Thanks!

burdiyan:
A block in IPFS is normally around 1-2 MB, and bigger files are split into chunks around this size, and then hashed. Even though reserving these multicodecs could be useful, they may be incompatible with how IPFS currently works.

These are my assumptions only though, I'm not a maintainer.

rvagg:
The 1-2 Mb limit is somewhat artificial and don't apply to most places that IPLD blocks are used; I think bitswap is the only place where block sizes are most constraintd. Plus it's not really a big concern for multicodecs because we don't assume that you're going to be passing them through IPFS systems or even using them in CIDs.

@drernie I think a pull request should be fine for these, what you say here sounds reasonable, one of us will probably have a look in a bit more detail just to sanity check that it's logical for them to have separate entries.

vmx:
> We would like to use a multihash for the new [Amazon S3 Checkums](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/), specifically their SHA-256 variant.

I've read the blog post. Do I understand correctly that the resulting hash of the Amazon S3 Checksum is equal to the on I'd calculate locally with a standard SHA-256 tool? If yes, what is your use case to have a different multicodec and not using the existing SHA-256 multihash identifier?

[...]

PR #343, opened 2024-02-15
#342 Amazon S3's parallel SHA2-256 with fixed 5/8MB blocks

drernie:
Registering prefixes for Amazon S3's parallel SHA2-256 with fixed 5MB or 8MB blocks
(the ones they actually implement in their SDK, and for which we are building a compatible implementation.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -510,8 +510,7 @@ poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,      Pose
 rdfc-1,                         ipld,           0xb403,         draft,      The result of canonicalizing an input according to RDFC-1.0 and then expressing its hash value as a multihash value.
 ssz,                            serialization,  0xb501,         draft,      SimpleSerialize (SSZ) serialization
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,      SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
-s3-sha2-256-5,                  multihash,      0xb515,         draft,      Amazon S3's parallel SHA2-256 with fixed 5MB blocks
-s3-sha2-256-8,                  multihash,      0xb518,         draft,      Amazon S3's parallel SHA2-256 with fixed 8MB blocks
+s3-sha2-256-boto,               multihash,      0xb511,         draft,      Boto3-compatible implementation Amazon S3's parallel SHA2-256 checksums 
 json-jcs,                       ipld,           0xb601,         draft,      The result of canonicalizing an input according to JCS - JSON Canonicalisation Scheme (RFC 8785)
 iscc,                           softhash,       0xcc01,         draft,      ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,      0xcert Asset Imprint (root hash)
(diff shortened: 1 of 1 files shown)


rvagg:
Discussion active in https://github.com/multiformats/multicodec/issues/342

table.csv
```diff
@@ -510,7 +510,7 @@ poseidon-bls12_381-a2-fc1-sc,   multihash,      0xb402,         draft,      Pose
 rdfc-1,                         ipld,           0xb403,         draft,      The result of canonicalizing an input according to RDFC-1.0 and then expressing its hash value as a multihash value.
 ssz,                            serialization,  0xb501,         draft,      SimpleSerialize (SSZ) serialization
 ssz-sha2-256-bmt,               multihash,      0xb502,         draft,      SSZ Merkle tree root using SHA2-256 as the hashing function and SSZ serialization for the block binary
-sha2-256-chunked,               multihash,      0xb510,         draft,      Hash of concatenated SHA2-256 digests of 8*2^n MiB source chunks; n = source_size / (10000 * 8MiB)
+sha2-256-chunked,               multihash,      0xb510,         draft,      Hash of concatenated SHA2-256 digests of 8*2^n MiB source chunks; n = log(source_size/(10^4 * 8MiB))
```
vmx, line 513: May I suggest (for being extra precise):

    ceil(log2(source_size / (10^4 * 8MiB)))

drernie: Ah, it is floor, so yes it would be good to be more precise :0)

vmx: Did I get my math wrong? (It can well be, it's hard to believe how much time I've spent trying to get this right ;) So let me do the math here again:

Let's say my input is 100GiB, hence `100*1024^3 = 107374182400` bytes. This means:

    log2((100*1024^3) / (10^4 * 8 * 1024^2))
    = 0.356143810225

 - `floor` would mean 0 => `8*2^0 = 8` MiB chunk size. `8MiB * 10^4 = 8*1024^2 * 10^4 = 83886080000`
 - `ceil` would mean 1 => `8*2^1 = 16` MiB chunk size `16MiB * 10^4 = 16*1024^2 * 10^4 = 167772160000`

    83886080000 < 107374182400 < 167772160000

So I would say it's ceil.

rvagg: Not going to object but I'm not really a fan of this description, too complex and the fact that you're having trouble figuring it out here means that the casual observer is going to have an almost impossible time (think of the case of someone who has a new hash function they want to register that sounds the same but they need to work out if it is).

Also it doesn't work for sizes under threshold (`10^4 * 8 MiB`) because it can go negative unless you're doing this in a strongly typed language. It really needs a branch for that case.

```js
> n = (sz) => Math.ceil(Math.log2(sz/((10**4)*(8*1024*1024))))
n(10000)
-23
```

Yeah, a `max(0, ...)` would fix it but yay more complexity!

Perhaps it could be fixed by speaking about the common (80000 MiB source) case and then talking how to scale beyond with the equation.

Don't want to drag this out though!

rvagg (approved):

drernie:
Thank you so much @rvagg !
Anyone else we need to ping to merge this?

table.csv
vmx: I just saw your comment after I merged it (as I saw your approval). In the squashed commit message i linked to the description document. I hope that's good enough.

I wonder if we should introduce yet another column with links to descriptions of the algorithm. But that would be a separate discussion/PR.

drernie:
🙏

PR #345, opened 2024-04-07
add Cryptid codecs

dhuseby:
This adds a number of multicodec values for various Cryptid projects ahead of the public release of the code. Most of the code is already available:
* [Multicid](https://github.com/cryptidtech/multicid)
  * Contains the VLAD implementation. [Spec](https://github.com/cryptidtech/provenance-specifications/blob/main/specifications/vlad.md)
* [Multisig](https://github.com/cryptidtech/multisig)
  * [Spec](https://github.com/cryptidtech/provenance-specifications/blob/main/specifications/multisig.md) 
* [Multikey](https://github.com/cryptidtech/multikey)
  * [Spec](https://github.com/cryptidtech/provenance-specifications/blob/main/specifications/multikey.md)
  *Contains the Nonce implementation as well. [Spec](https://github.com/cryptidtech/provenance-specifications/blob/main/specifications/nonce.md)

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -5,6 +5,9 @@ cidv2,                          cid,            0x02,           draft,      CIDv
 cidv3,                          cid,            0x03,           draft,      CIDv3
 ip4,                            multiaddr,      0x04,           permanent,
 tcp,                            multiaddr,      0x06,           permanent,
+vlad,                           cid,            0x07,           draft,      Verifiable Long-lived ADdress
+provenance-log,                 ipld,           0x08,           draft,      Verifiable and permissioned append only log
+provenance-log-entry,           ipld,           0x09,           draft,      Verifiable and permissioned append only log
 sha1,                           multihash,      0x11,           permanent,
 sha2-256,                       multihash,      0x12,           permanent,
 sha2-512,                       multihash,      0x13,           permanent,
@@ -36,6 +39,9 @@ dns,                            multiaddr,      0x35,           permanent,
 dns4,                           multiaddr,      0x36,           permanent,
 dns6,                           multiaddr,      0x37,           permanent,
 dnsaddr,                        multiaddr,      0x38,           permanent,
+multisig,                       multiformat,    0x39,           draft,      Digital signature multiformat
+multikey,                       multiformat,    0x3a,           draft,      Encryption key multiformat 
+nonce,                          key,            0x3b,           draft,      Nonce random value
 protobuf,                       serialization,  0x50,           draft,      Protocol Buffers
 cbor,                           ipld,           0x51,           permanent,  CBOR
 raw,                            ipld,           0x55,           permanent,  raw binary
@@ -71,6 +77,7 @@ aes-192,                        key,            0xa1,           draft,      192-
 aes-256,                        key,            0xa2,           draft,      256-bit AES symmetric key
 chacha-128,                     key,            0xa3,           draft,      128-bit ChaCha symmetric key
 chacha-256,                     key,            0xa4,           draft,      256-bit ChaCha symmetric key
+chacha20-poly1305,              multikey,       0xa5,           draft,      ChaCha20_Poly1305 encryption scheme
 bitcoin-block,                  ipld,           0xb0,           permanent,  Bitcoin Block
 bitcoin-tx,                     ipld,           0xb1,           permanent,  Bitcoin Tx
 bitcoin-witness-commitment,     ipld,           0xb2,           permanent,  Bitcoin Witness Commitment
@@ -176,6 +183,25 @@ p521-priv,                      key,            0x1308,         draft,      P-52
 bls12_381-g1-priv,              key,            0x1309,         draft,      BLS12-381 G1 private key
 bls12_381-g2-priv,              key,            0x130a,         draft,      BLS12-381 G2 private key
 bls12_381-g1g2-priv,            key,            0x130b,         draft,      BLS12-381 G1 and G2 private key
+bls12_381-g1-pub-share,         key,            0x130c,         draft,      BLS12-381 G1 public key share
+bls12_381-g2-pub-share,         key,            0x130d,         draft,      BLS12-381 G2 public key share
+bls12_381-g1-priv-share,        key,            0x130e,         draft,      BLS12-381 G1 private key share
+bls12_381-g2-priv-share,        key,            0x130f,         draft,      BLS12-381 G2 private key share
+lamport-sha3-512-pub,           key,            0x1a14,         draft,      Lamport public key based on SHA3-512
+lamport-sha3-384-pub,           key,            0x1a15,         draft,      Lamport public key based on SHA3-384
+lamport-sha3-256-pub,           key,            0x1a16,         draft,      Lamport public key based on SHA3-256
+lamport-sha3-512-priv,          key,            0x1a24,         draft,      Lamport private key based on SHA3-512
+lamport-sha3-384-priv,          key,            0x1a25,         draft,      Lamport private key based on SHA3-384
+lamport-sha3-256-priv,          key,            0x1a26,         draft,      Lamport private key based on SHA3-256
+lamport-sha3-512-priv-share,    key,            0x1a34,         draft,      Lamport private key share based on SHA3-512 and split with Shamir gf256
+lamport-sha3-384-priv-share,    key,            0x1a35,         draft,      Lamport private key share based on SHA3-384 and split with Shamir gf256
+lamport-sha3-256-priv-share,    key,            0x1a36,         draft,      Lamport private key share based on SHA3-256 and split with Shamir gf256
+lamport-sha3-512-sig,           multisig,       0x1a44,         draft,      Lamport signature based on SHA3-512
+lamport-sha3-384-sig,           multisig,       0x1a45,         draft,      Lamport signature based on SHA3-384
+lamport-sha3-256-sig,           multisig,       0x1a46,         draft,      Lamport signature based on SHA3-256
+lamport-sha3-512-sig-share,     multisig,       0x1a54,         draft,      Lamport signature share based on SHA3-512 and split with Shamir gf256
+lamport-sha3-384-sig-share,     multisig,       0x1a55,         draft,      Lamport signature share based on SHA3-384 and split with Shamir gf256
+lamport-sha3-256-sig-share,     multisig,       0x1a56,         draft,      Lamport signature share based on SHA3-256 and split with Shamir gf256
 kangarootwelve,                 multihash,      0x1d01,         draft,      KangarooTwelve is an extendable-output hash function based on Keccak-p
 aes-gcm-256,                    encryption,     0x2000,         draft,      AES Galois/Counter Mode with 256-bit key and 12-byte IV
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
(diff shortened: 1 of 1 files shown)


dhuseby:
The new Multisig and Multikey specs and implementations strictly follow:
* They MUST be in-band (with the value); not out-of-band (in context).
* They MUST avoid lock-in and promote extensibility.
* They MUST be compact and have a binary-packed representation.
* They MUST have a human-readable representation.

This cannot be said for the existing Varsig spec. It relies heavily on context to know how to decode the bytes of the digital signature. More importantly, if your implementation doesn't support a specific algorithm, there is no way to know how many bytes to skip over in the stream to skip over the Varsig. Multisig fixes this as an explicit goal.

rvagg:
Apologies for our slowness on this @dhuseby, both Volker and I are a bit stretched and you've given us a lot to review with your PRs and the cost of context switching to try and get our heads around all the new stuff in here isn't small.

Firstly, the one we tend to be most concerned about is anything that adds `ipld` - i.e. anything that might qualify for going into the "codec" portion of a CID.

1. If it's not intended to live in a CID then `ipld` is probably the wrong choice.
2. If it follows the same decoding and encoding process as an existing codec, then adding a new entry is probably not appropriate. It should be a new encoding format that's not already covered. We discourage using codec codes as a signalling mechanism for anything other than "how do I turn these bytes into the IPLD data model and back again".
3. If decoding bytes using this codec doesn't yield any "links" (concretely: CIDs), such that it can contribute to a larger graph, then `serialization` may be a better choice.

The new multi* entries will also require some additional thought and consideration. And due to the changes of the tags for existing entries we may need to pull in some folks who rely on those.

dhuseby:
@rvagg 

> If decoding bytes using this codec doesn't yield any "links" (concretely: CIDs), such that it can contribute to a larger graph, then serialization may be a better choice.

Ah...ok. These are a new IPLD data structure but I think you're right, `serialization` is probably correct here.

> The new multi* entries will also require some additional thought and consideration. And due to the changes of the tags for existing entries we may need to pull in some folks who rely on those.

This is a fair concern. I think to be safe, I'll just duplicate the overlap and assign new numbers.

Let me fix this.

dhuseby:
Thank you for taking time to review this. I changed the provenance-log and provenance-log-entry to be `serialization`. I changed a VLAD from `cid` to `vlad` since it is an identifier but not a CID. I restored the `varsig` entries to their original state and moved all of the `multisig` signature codecs to unique names and values to avoid collision. This should streamline the review and eliminate any push-back from repurposing.

dhuseby:
I also just corrected the `es384-msig` name error. I filed an issue for the same name error for the varsig: #348

dhuseby:
Also just cleaned up the `es521-msig` name error. I filed an issue for the same name error for varsig: #349

vmx:
I only had a quick look. We are generally pretty cautious adding anything into the number space which is encoded into a single byte as it is so limited (also some things that are already there shouldn't really, but that's hard to change). So if you would move your additions to a 2 bytes or higher section, that would be great.

dhuseby:
> So if you would move your additions to a 2 bytes or higher section, that would be great.

I can do that, but what about the `multikey`, `multisig`, `nonce`, and the `chacha20-poly1305`? I put all of these in the single byte space because they are fundamental types (or soon will be). I put `chacha20-poly1305` to be next to the existing `chacha` codecs.

I'm happy to move things elsewhere but I had reasons for putting these in the single byte space.

vmx:
> (or soon will be).

That's the point. At the moment we hardly add anything to the range of single bytes, the idea is to put things there that have wide adoption/are standardized. I know that it's annoying if it really gets used, that later on you don't really change to a different code, but that space is just so limited.

burdiyan:
FWIW, the word Multikey is currently being used in some of the upcoming W3C specs related to DID and Verifiable Credentials: https://www.w3.org/TR/vc-data-integrity/#multikey.

dhuseby:
My multikey predates the W3C use. There's no mention of multikey in this file yet. Plus the verifiable credentials spec has failed to gain adoption in key political and business areas.

table.csv
```diff
@@ -517,9 +544,10 @@ json-jcs,                       ipld,           0xb601,         draft,      The
 iscc,                           softhash,       0xcc01,         draft,      ISCC (International Standard Content Code) - similarity preserving hash
 zeroxcert-imprint-256,          zeroxcert,      0xce11,         draft,      0xcert Asset Imprint (root hash)
 nonstandard-sig,                varsig,         0xd000,         deprecated, Namespace for all not yet standard signature algorithms
+bcrypt-pbkdf,                   multihash,      0xd00d,         draft,      Bcrypt-PBKDF key derivation function
 es256k,                         varsig,         0xd0e7,         draft,      ES256K Siganture Algorithm (secp256k1)
-bls-12381-g1-sig,               varsig,         0xd0ea,         draft,      G1 signature for BLS-12381-G2
-bls-12381-g2-sig,               varsig,         0xd0eb,         draft,      G2 signature for BLS-12381-G1
+bls12_381-g1-sig,               varsig,         0xd0ea,         draft,      G1 signature for BLS-12381-G2
+bls12_381-g2-sig,               varsig,         0xd0eb,         draft,      G2 signature for BLS-12381-G1
```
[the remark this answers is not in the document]
dhuseby: reverted

vmx (approved):
Those are lot of codecs, but as all of them are at least in the 2 byte range I don't see a reason not to accept them. We surely want people to use the multicodec table and experiment with things. Hence I approve it.

But before merging I'd like to see another approval from someone else, as it isn't just me making decisions.

table.csv
```diff
@@ -72,6 +72,7 @@ aes-192,                        key,            0xa1,           draft,      192-
 aes-256,                        key,            0xa2,           draft,      256-bit AES symmetric key
 chacha-128,                     key,            0xa3,           draft,      128-bit ChaCha symmetric key
 chacha-256,                     key,            0xa4,           draft,      256-bit ChaCha symmetric key
+chacha20-poly1305,              multikey,       0xa5,           draft,      ChaCha20_Poly1305 encryption scheme
```
rvagg, line 75: One minor flag here - this is multi-byte so not a big deal, but adding this here doesn't leave any space for the plethora of authenticated encryption algorithms that it belongs with, I can easily imagine someone showing up and wanting to add a ton more but there being no space here. Maybe this belongs somewhere else where there's a bit more space to expand the AE list.

"multikey" seems a little limiting for this, but I don't have a better suggestion and I can't think of another way this might be consumed outside whatever you're doing.

rvagg:
I think I'm OK with this, it's just a lot of new tags being added, including single-entry ones: vlag, nonce, multikey. I'd be more comfortable if we didn't encourage more expansion of tags. But I don't have a hard reason to object.

table.csv
dhuseby: The multikey part is because Multikey objects can contain a chacha20-poly1305 symmetric key and tag. Multikeys can also contain an encrypted secret key that is encrypted using chacha20-poly1305.

I'm happy to move this to a 2-byte value.

dhuseby:
> including single-entry ones: vlag, nonce, multikey

At the risk of bikeshedding... vlad and cid tags should just be "identifier" or something generic like that. multikey and key could be combine into "key-like" or something generic like that. But that would require re-tagging a bunch of values which I'm sure is way worse of an outcome. The tags are all fundamental types that fit with the establish pattern IMO.

Thanks for the time you spent reviewing these.

table.csv
rvagg: It already encodes to a 2-byte varint so it's not a major blocker, it just doesn't allow room for additional AE schemes in this block, the only way it really fits is because it's next to chacha-* but it's actually quite different than plain chacha. Not a big deal, just trying to think ahead.

dhuseby: Want me to move it? I'm happy to do that.

rvagg: hey, how about we just bump it to `0xa00`, there's a huge block up there if we end up needing it, and 0xa00 encodes to the same number of varint bytes (2) as `0xa5`.

dhuseby:
Thanks! Can we land this?

table.csv
dhuseby: ok, let me do that

dhuseby:
I think all concerns have been addressed.

rvagg:
Sorry for the delays Dave, we're not well set up for dealing with complex requests like this.

vmx:
@dhuseby Thanks for your patience and all the changes compared to the original PR!

bumblefudge:
@rvagg @vmx Sorry for the late appearance, this crossed my inbox last week but I was at TPAC on a borrowed laptop and couldn't get here in time.

> My multikey predates the W3C use

This kinda feels like a "trust me bro" assertion.  The "W3C use" was based on early versions of juan benet's eternally-[unfinished/unpublished spec from 2016](https://github.com/ipfs/specs/issues/58) which seems consistent with references in the original commits in this org. In fact, I myself [removed `multikey` from the top-level readme in a 2023 PR](https://github.com/multiformats/multiformats/pull/65/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L20) exactly because I found discrepancies in various [proposed multikey variants](https://github.com/multiformats/multicodec/pull/125#discussion_r363243492) in the history of this org but no spec more canonical than JB's original issue (which matches _earlier versions_ of W3C usage, fwiw) to discriminate between the variants. The new spec seems perfectly defensible but I would consider giving it a suffix or renaming it to avoid confusion with the older `multikey` and its variants. If it doesn't get renamed here in this org, it might just cause headaches if this ever goes to IANA.

FWIW W3C DIDWG and VCWG do not use the term "multikey" anywhere, nor do they use by another name a similar CID-like multiformat; what they actually use (and specify internally) is a multibased raw-codec ULEB128 of public key bytes with the w3c-namespaced RDF/JSON-LD semantic [`publicKeyMultibase`](https://www.w3.org/TR/did-core/#dfn-publickeymultibase). Which is to say, the potential for confusion is, luckily, not as significant for the W3C term as for the multiformats' homegrown historical precedents. AFAIK none of the representatives from Maersk, GS1, or the Singaporean Trade Commission were lying about using `publicKeyMultibase` in prod last week at W3C TPAC?

Similarly, I believe there is some [multisig work ongoing elsewhere in the PL cinematic universe](https://github.com/ucan-wg/receipt?tab=readme-ov-file#depends-on) that I referred OP to when they asked about this a year ago, and never heard back about. That one goes by `varsig` rather than multisig and I'm not sure it's going to production or not, but could also be a source of avoidable confusion. Probably nothing but odd it wasn't mentioned in the PR.

dhuseby:
@bumblefudge I don't care what it is called. If "multikey" is a hotly contested name then let's go with "polykey" or "unikey" or "whateverkey" or "skeletonkey". I officially do not care. I just need these changes to get landed. So in my best effort to end the yak shaving and to celebrate Halloween/Diwali and our triumph of light over darkness/enlightenment over ignorance, I propose the use of "skeletonkey" to alleviate the concern.

dhuseby:
If Rust allowed for unicode type names, I'd even go so far as to call it "💀 key"

dhuseby:
> @dhuseby Thanks for your patience and all the changes compared to the original PR!

No worries! This is ~~Thunderdome~~ open source. 🪓

bumblefudge:
oh totally the sigil is yours! sorry it didn't land sooner.

dhuseby:
Thank you!

PR #347, opened 2024-04-24
add `iroh-hashseq` codec

cameronfyfe:
This adds a codec for hash sequence blobs used by Iroh.

Docs describing hash sequence blobs: https://iroh.computer/docs/layers/blobs#collections


cameronfyfe:
Thanks for the quick response.

Yes, my mental map of the codec was mapping to `list<link>` with pre-defined hash/codec as you described.

I can see the argument for making it more general. What about just `blake3-hashseq`? I could see any variation of `{blake3|blake3-32|blake3-256}{hashseq|seq}` making sense but that's my first choice.

vmx:
`blake3-hashseq` sounds good to me. @rvagg what do you think?

rvagg (approved):
sgtm

vmx (approved):

PR #352, opened 2024-06-05
Add http-path

MarcoPolo:
The multicodec corollary to https://github.com/multiformats/multiaddr/pull/164

closes #324


rvagg (approved):
👌 nice to get this formalised

rvagg:
History here is that ipni introduced this originally as `httpath` and there's been a multi-year discussion about getting it formalised; I think ipni still uses `httpath` so may have to do some bridging to `http-path`.

We probably should ping @willscott and @gammazero about this I guess. It'll close the one I opened but was objected to @ https://github.com/multiformats/multicodec/pull/324

PR #353, opened 2024-06-10
Codecs for ENS Contenthash: URI [0xF2] and Data URL [0xF3]

adraffy:
[ENS](https://docs.ens.domains) (Ethereum Name Service) encodes [`contenthash()`](https://docs.ens.domains/ensip/7) using multicodec.  The purpose of a `contenthash()` is to describe the web contents for a corresponding ENS name.

Currently, ENS supports IPFS, IPNS, Swarm, Arweave, Onion, etc.  

Example using IPFS:
*  ENS: [`vitalik.eth`](https://app.ens.domains/vitalik.eth) 
*  ENS Web Gateway using Contenthash: [`https://vitalik.eth.limo/`](https://vitalik.eth.limo/) 
* `contenthash()` = `0xe30101701220484da2f7f497cac307e2026282263630b8dd4c448c3436470f5b850b432ba868` 
*  Decoded URL: [`ipfs://k2jmtxt5zh5vu5y8r7em2che3d4ghyftfr6h1yofdhibxai88k1wj5uw`](https://adraffy.github.io/cid.js/test/demo.html#0xe30101701220484da2f7f497cac307e2026282263630b8dd4c448c3436470f5b850b432ba868)
* `0xE3` correspond to multicodec `ipfs`, the following bytes are a CIDv0

We would like to support the following (2) new codecs:

* `0xF2` — **URI**
     * Encoded: `0xf268747470733a2f2f656e732e646f6d61696e732f`
     * Format: `<codec><url: utf8-string>`
     * Decoded URL: `https://ens.domains/` 
* `0xF3` — **Data URL**
     * Encoded: `0xf309746578742f68746d6c3c68746d6c3e68656c6c6f3c2f68746d6c3e`
     * Format: `<len(mime): uint8><mime: ascii-string><data: uint8[]>`
     * Decoded:
          * `mime` = `text/html` (9ch)
          * `data` = `<html>hello</html>` (encoding depends on mime)

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -100,6 +100,8 @@ bls12_381-g1g2-pub,             key,            0xee,           draft,      BLS1
 sr25519-pub,                    key,            0xef,           draft,      Sr25519 public key
 dash-block,                     ipld,           0xf0,           draft,      Dash Block
 dash-tx,                        ipld,           0xf1,           draft,      Dash Tx
+uri,                            namespace,      0xf2,           draft,      URI: <url: utf8-string>
+data-url,                       data,           0xf3,           draft,      data URL: <len(mime): uint8><mime: ascii-string><data: uint8[]>
 swarm-manifest,                 ipld,           0xfa,           draft,      Swarm Manifest
 swarm-feed,                     ipld,           0xfb,           draft,      Swarm Feed
 beeson,                         ipld,           0xfc,           draft,      Swarm BeeSon


rvagg:
I think this seems reasonable, though novel. I'm not so sure about introducing a new tag, `data` for this though. Would `namespace` as well for that be OK? Even that doesn't map super cleanly onto what you're doing here.

Do you think you'll want more of these into the future? I wonder if we can't figure out a better tag whether this should just be an entirely new classification.

@vmx, what do you think?

vmx:
I wonder if URI could use a Multiaddress instead. Would that be an option (I know to little about the Eth/ENS ecosystem).

adraffy:
`namespace` works.  I'd be happy to change it to whatever you suggest.

IMO, the closest codec is `json` which oddly uses tag `ipld`.

I picked `data` as unlike most codecs, `data-uri` is both a codec and the data itself. 

---

I think `multiaddr` for `uri` suggests too much internal encoding, as we want something maximally general (a literal UTF-8 string) where the content is ultimately validated by the client (since URL standards are ever-evolving)

vmx:
> I think `tag:multiaddr` for `uri` suggests too much internal encoding, as we want something maximally general (a literal UTF-8 string) where the content is ultimately validated by the client (since URL standards are ever-evolving)

Keeping it simple makes sense.

aschmahmann:
IIUC this is related to https://discuss.ens.domains/t/draft-ensip-17-datauri-format-in-contenthash/18048/28 and https://github.com/ensdomains/docs/pull/165.

Apologies for the long text, I'm going to be OOO for a couple days and wanted to make sure to leave some context. cc @lidel who has been involved in the ENS work and interop here since long before me 😅.

TLDR:
- As per usual, unless a codec makes very little sense, is duplicative, or seems to trivially open the door for a whole bunch more codecs I'm generally +1 on applications - although sometimes I recommend moving to a higher byte range in the table 
- DataURL in particular seems like something the ENS and IPFS communities could work on, if the mime-type issue is causing them big enough problems that identity raw CIDs are insufficient then it seems like that'll happen for data that's too big to reasonably use a DataURL for.

Some thoughts:

## URI

> I wonder if URI could use a Multiaddress instead

Probably not multiaddress itself, but harmonization with something like multipath https://github.com/multiformats/multiformats/pull/55 would likely make this work and be pretty sensible. It would likely also let us use the `0x2f` as an escape hatch for people generally wanting to use/experiment with strings rather than code numbers which is what this roughly does (otherwise, the codes like for http could potentially be used instead).

FWIW libp2p has recently proposed going the other way as well (i.e. representing multiaddrs as URIs https://github.com/multiformats/multiaddr/pull/171).

I don't in principle have an objection to a URI based namespace, the two byte range is probably fine although URIs could probably tolerate even three due to the size of the data.

Perhaps more of an ENS-related comment, but want to call out:
1. There is some redundancy here because for any namespace (IPFS, Swarm, etc.) you could encode under the URI namespace or under their individual namespace. Not necessarily a big issue here, but certainly a change implementations will need to take care of
2. Related to ^ it seems like this could have always been the case, I'm not sure the historical context here but probably worth validating with folks who did this in earlier rounds that this makes sense. Totally fair + reasonable to say we want to save some bytes with known namespaces and then have the utf-8 URI escape hatch (although I don't know if "contenthash" is a reasonable name for this kind of thing 🙃).

## Data URL

Seems fine, although maybe the three byte range (along with arweave, skynet, etc.) makes more sense here given these will likely be larger anyhow.

A few comments / thoughts:
1. Given the above technically this already works as a Data URI, right https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs, right? If so, I assume the idea is to preserve space by not needing to do base64 encoding.
2. While saving some bytes here seems fine. This seems non-optimal in that it both isn't as compact as it could be (e.g. mime types are still expressed in text), not flexible enough to include any other metadata, and we couldn't work around it within the existing namespaces (the URI namespace adds a sort of escape hatch here as long as you assume names won't collide).
  - In my bias as someone who works on the IPFS project IMO this could've/should've been resolved by having the tooling for this either in IPFS (either in UnixFS, CID, or another IPLD format), and this seems like as good a time as any to resolve it independently of what happens in this PR (although it may justify bumping to 3 bytes)
     - A CID with the identity multihash and raw codec (or sometimes codecs like JSON or CBOR) would've been sufficient except for the need of a mimetype
     - Technically this could be resolved in a few different ways, one is https://github.com/ipfs/specs/issues/257, note: the latest request here came from the ENS community as well so definitely seems like a good opportunity to chat anyhow
     - Given the very large number of ENS `contenthash` records that are IPFS-based this seems like something we could/should fix or the hack within ENS (whether in ENS or the "contenthash" namespace could fix either)
     - I understand this isn't really the place for an ENSIP comment and with my "multiformats hat" I don't have objection, but if you want to chat would definitely be happy to

PR #357, opened 2024-10-01
Merkle Reference Code

Gozala:
This PR proposes addition of the code for [merkle-references described in the specification](https://github.com/storacha/RFC/blob/feat/merkle-structure/rfc/merkle-structure.md). Allocating a code would enable us to format those references as IPLD links and integrate them into rest of the IPLD ecosystem


Gozala:
I could also use some feedback / recommendations on the best way to integrate them into the ecosystem as I've being going back and forth on this. Merkle reference should be considered as a hashing algorithm defined for all of the  [IPLD types](https://ipld.io/docs/data-model/kinds/) (as opposed to just bytes). However similar to CIDs it could use different underlying hashing algorithms in addressing (although same hashing algorithm across the complete DAG is implied).

For this reason merkle-reference is proposed as a standalone multiformat that has following structure

```
<merkle-reference> ::= <merkle-reference-multicodec><content-multihash>
# or, expanded:
<merkle-reference> ::= <`0x07`, the code for merkle-reference><multihash of merkle-folded data>
```

There is a nice duality to this. As you can also describe `0x07` as IPLD codec, if you describe merkle-folding process as lossy encoding. Therefor, if you prefix  merkle-reference with `0x01` varint you'll end up with valid CIDv1 that could be integrated into rest of the IPLD ecosystem.

alanshaw:
I like the idea. I don't know the process to have codes accepted here, do we require at least one implementation?

Gozala:
> I like the idea. I don't know the process to have codes accepted here, do we require at least one implementation?

I have an implementation that currently lives here https://github.com/Gozala/merkle-reference

bumblefudge:
> I like the idea. I don't know the process to have codes accepted here, do we require at least one implementation?

See [Robin's process doc](https://github.com/multiformats/multiformats/blob/master/contributing.md#multiformats-registrations) - `Provide evidence that the encoding is supported in at least two production implementations.` is required for DRAFT status, and this would also be a requirement if the registry were administered at IANA or at W3C's new registry process. Since we're in the extremely scarce single-byte range, it might make more sense to wait until this is a little more developed/further along before robbing future-multiformats of one slot needed for future CID and/or other structured "third layer" tags? 

Another thing that would help would be if the current spec were formatted as a complete, linter-passing IETF internet-draft or W3C CG respec doc with test-vectors and all that, rather than an open PR with unresolved comments on the web3.storage IP process repo... not technically a requirement at DRAFT status but maybe one that would help allay concerns of single-byte squatting from the least-generous possible readers 😉

Gozala:
> See [Robin's process doc](https://github.com/multiformats/multiformats/blob/master/contributing.md#multiformats-registrations) - `Provide evidence that the encoding is supported in at least two production implementations.` is required for DRAFT status, and this would also be a requirement if the registry were administered at IANA or at W3C's new registry process. 

Ah I was not aware of the new process, thanks for pointing it out.

> Since we're in the extremely scarce single-byte range, it might make more sense to wait until this is a little more developed/further along before robbing future-multiformats of one slot needed for future CID and/or other structured "third layer" tags?

Sounds reasonable, yet seems like a double standard when I see 

https://github.com/multiformats/multicodec/blob/352d05ad430713088e867216152725f581387bc8/table.csv#L4-L5

CIDv2 had being discussed forever and I would be very surprised if there are multiple implementations two production implementations. I have not even heard of CIDv3 probably something new that happened since I fell of the inter planetary space 😅 I won't even mention that most of the codes in that table would fail to meet new criteria.

As of second implementation, there is one in development in Rust and I can update the thread here when it's ready.

> Another thing that would help would be if the current spec were formatted as a complete, linter-passing IETF internet-draft ([example](https://learningproof.github.io/tid-i-d/draft-goldman-tid.html) from a multibase registration in progress) or W3C CG respec doc with test-vectors and all that, rather than an open PR with unresolved comments on the web3.storage IP process repo... not technically a requirement at DRAFT status but maybe one that would help allay concerns of single-byte squatting from the least-generous possible readers 😉

Most up to date spec lives here https://github.com/Gozala/merkle-reference/blob/main/docs/spec.md.  There is also interactive version https://observablehq.com/@gozala/merkle-references that anyone can test with various data sets.

Some [test fixtures are available here](https://github.com/Gozala/merkle-reference/blob/c6c5127a4841c0cdb06b98977f1484c4b06b2675/test/lib.spec.js#L4-L45) and they'll likely move into more portable form once Rust implementation is there.

Trying to reformat it into IETF / W3C spec format is plausible, but as one man show I got to be pragmatic with where I spend time and it seemed a lot more reasonable to budget it after code is in the table as opposed to before.

Gozala:
I should mention that presence on multicodec table is nice to have mostly for backwards compatibility with IPLD addressing scheme. In practice I don't expect multiformat prefixes to be used beyond bridging with legacy (IPLD) system.

bumblefudge:
> Sounds reasonable, yet seems like a double standard when I see
> 
> https://github.com/multiformats/multicodec/blob/352d05ad430713088e867216152725f581387bc8/table.csv#L4-L5

Those two are a special case of "reserved for future use", as far as I know, and there is no definite timeline on CIDv2 or CIDv3, I think the single-bytes are just being held in reserve for future generations.  And in total transparency, much of what's in the multicodec table today, including many "final" registrations, particularly those things with a casual/light spec and no second implementation, would need to be summarily demoted to "vendor", "experimental", or "reserved" status if the table moved into IANA or W3C governance.

> Trying to reformat it into IETF / W3C spec format is plausible, but as one man show I got to be pragmatic with where I spend time and it seemed a lot more reasonable to budget it after code is in the table as opposed to before.

Oh, of course, if I'm still engaged in this project when you get the running code and informal spec over the line, I can help with formalizing the spec, and if I'm not, hopefully someone at IPFS Foundation or Shipyard can help instead. Sprucing up a spec is a laborious chore you won't have to do alone if there is interest and other people are interested in using this alternative to conventional CIDv1s!

burdiyan:
This new approach of hashing data structures seems pretty interesting. Actually, it's probably the closest thing to the original vision of IPLD out there. I wonder whether it can be made compatible with existing IPFS data though.

rvagg:
Could we move this to the two-byte range, at least for now? I'm concerned that without a community of people developing this or some kind of institutional backing that it's not going to get enough traction to go anywhere beyond an interesting idea. I feel bad saying this, but PL's nucleation process has meant a lot of good ideas like this have had to be relegated to the personal experiments of individuals. Squatting `0x7` for something that's not baked in to someone's product or has an excited community of developers is hard to justify when the precious 1-byte range is the thing we're the most protective of in the table these days.

Gozala:
> This new approach of hashing data structures seems pretty interesting. Actually, it's probably the closest thing to the original vision of IPLD out there. I wonder whether it can be made compatible with existing IPFS data though.

There is a [integration section](https://github.com/Gozala/merkle-reference?tab=readme-ov-file#ipld-integration) in the repo that describes how compatibility with IPLD is currently managed.

I have used it successfully with various IPLD codecs that treat merkle-references as IPLD Links a.k.a CIDs. Happy to discuss this more, but lets do it n the linked repo instead to reduce a noise here.

Gozala:
> Could we move this to the two-byte range, at least for now? I'm concerned that without a community of people developing this or some kind of institutional backing that it's not going to get enough traction to go anywhere beyond an interesting idea. 

It is being used by a startup I'm currently employed by. I believe @hannahhoward also was looking at some point into using this for indexing blockchains, although I do not know if things have change there.

> I feel bad saying this, but PL's nucleation process has meant a lot of good ideas like this have had to be relegated to the personal experiments of individuals. Squatting `0x7` for something that's not baked in to someone's product or has an excited community of developers is hard to justify when the precious 1-byte range is the thing we're the most protective of in the table these days.

I understand the rational, although at this point I don't find much value in pursuing that. In our current use we already treat merkle-references as multihash to avoid 1 byte CID overhead, yet we still pay 3 byte overhead. 

If there low interest to justify code allocation that probably implies there is equally low probability of collision so I'd rather revisit this when and if interest / probability is higher.

Gozala:
As a side note it might be a good idea to switch to hierarchical tables like we did with [varsig](https://github.com/ChainAgnostic/varsig) that way context specific tables can exist without every single format having to compete for the canonical code.

In such world merkle-reference would have being registered as a IPLD codec which would have had much smaller table.

burdiyan:
> As a side note it might be a good idea to switch to hierarchical tables like we did with [varsig](https://github.com/ChainAgnostic/varsig) that way domain specific tables can exist without every single format having to compete for the canonical code.
> 
> In such world merkle-reference would have being registered as a IPLD codec which would have had much smaller table.

So agree with this. Not sure about hierarchical, but namespaces could be very useful.

bumblefudge:
What's the status on varsig btw? see the discussion in #345

burdiyan:
Just found out that Amazon's ION encoding format has does something similar with their ion-hash: https://amazon-ion.github.io/ion-hash/.

hannahhoward:
Since I got tagged I remain super interested in merkle references and would love it if we could just merge it. I would definitely prefer a 2 byte merge over a 1 byte endless git thread.

hannahhoward:
@Gozala have you perf tested the rust impl? I'm honestly curious cause I see CPU usage as the biggest ??? in this approach which I'm otherwise super jazzed about.

Gozala:
> @Gozala have you perf tested the rust impl? I'm honestly curious cause I see CPU usage as the biggest ??? in this approach which I'm otherwise super jazzed about.

I'm afraid rust impl moved to back-burner. Although some recent changes in JS one should improve things a bit because it makes use of caches.

I think better implementation should probably take advantage of parallelizability of the algorithm and do something along the lines of what [fleek did with blake3](https://blog.fleek.network/post/fleek-network-blake3-case-study/) but in our use cases merkle-references had not being bottleneck so no time was spend to optimize it.

PR #358, opened 2024-10-08
Add ML-KEM public keys as specified by FIPS 203

amlord:
Add ML-KEM public keys, as specified by FIPS-203: https://csrc.nist.gov/pubs/fips/203/final


vmx (approved):
I haven't really looked into this. But it's in the two bytes range, so why not merging it. Though I'd like another approval from someone else before merging.

rvagg (approved):

PR #366, opened 2024-12-13
dms3 draft codec request

yozgatsi:
please consider this request to register a public codec for an early stage project. \thank you

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -585,3 +585,4 @@ es384-msig,                     multisig,       0xd01309,       draft,      ECDS
 es521-msig,                     multisig,       0xd0130a,       draft,      ECDSA P-521 Signature as Multisig
 rs256-msig,                     multisig,       0xd0130b,       draft,      RS256 Signature as Multisig
 scion,                          multiaddr,      0xd02000,       draft,      SCION Internet architecture
+dms3,                           multiaddr,      0xd03000,       draft,      dms3


rvagg:
Would you be able to provide a little bit more information for the description field? And are you expecting to be needing a lot more entries around this or can I suggest you move it down among the other numbers just to avoid stretching the far end of the spectrum unnecessarily. Or is there something about this number that is a nice fit for your project?

yozgatsi:
DMS3 = Decentralized Multi-State-Space Search (IR) platform on a p2p network. Nodes will leverage the technology, but form a separate network from that of IPFS/Libp2p. I am fine fine to move the number to compact the codec space. I don't expect to need a large block of numbers, and it's difficult to anticipate  all my needs at this time. If you allowed block reservations, an order of 10 may be appropriate.

yozgatsi:
perhaps Dms3Key would be a more appropriate name for my requested codec, as its use is similar to that of Libp2pKey.

table.csv
```diff
@@ -585,3 +585,4 @@ es384-msig,                     multisig,       0xd01309,       draft,      ECDS
 es521-msig,                     multisig,       0xd0130a,       draft,      ECDSA P-521 Signature as Multisig
 rs256-msig,                     multisig,       0xd0130b,       draft,      RS256 Signature as Multisig
 scion,                          multiaddr,      0xd02000,       draft,      SCION Internet architecture
+dms3,                           multiaddr,      0xd03000,       draft,      dms3
```
rvagg, line 588:
```suggestion
dms3,                           multiaddr,      0xd03000,       draft,      Decentralized Multi-State-Space Search
```

rvagg:
There's a ton of space above `0x100000`, whole blocks of the first digit in those are missing, there's no good reason to jump to `d`, you could even just take `0x100000` if you want. I'd just prefer we not stretch the space out further if there's not a good reason to. If it spelled something or the number had some relevance for you then that'd be enough, otherwise, just choose a lower number and you're good. If you pick a number with some space after it then you're welcome to come back later and allocate more. Particularly in this end of the number space, people are reluctant to bump right up against another un-related entry so you should be fine if you choose a block in that end of the table.

yozgatsi:
Hi Rod, thank you for the suggestion, works for me. Please let me know if you can apply the following changes. \Tavit

dms3,                           multiaddr,      0x100000,       draft,      Decentralized Multi-State-Space Search
dms3-key,                    ipld,               0x100001,       draft,      DMS3 Public Key

rvagg:
@yozgatsi please update this branch with the changes or, if you have trouble doing that feel free to close this and open a new pull request with it.

yozgatsi:
seem I got it right on the second try to modify the PR...

table.csv
```diff
@@ -557,6 +557,8 @@ eip-191,                        varsig,         0xd191,         draft,      EIP-
 jwk_jcs-pub,                    key,            0xeb51,         draft,      JSON object containing only the required members of a JWK (RFC 7518 and RFC 7517) representing the public key. Serialisation based on JCS (RFC 8785)
 fil-commitment-unsealed,        filecoin,       0xf101,         permanent,  Filecoin piece or sector data commitment merkle node/root (CommP & CommD)
 fil-commitment-sealed,          filecoin,       0xf102,         permanent,  Filecoin sector data commitment merkle node/root - sealed and replicated (CommR)
+dms3,                           multiaddr,      0x100000,       draft,      Decentralized Multi-State-Space Search
+dms3-key,                       ipld,           0x100001,       draft,      DMS3 Public Key
```
rvagg, line 561:
```suggestion
dms3-key,                        key,           0x100001,       draft,      Decentralized Multi-State-Space Search Public Key
```

unfortunately you can't use `ipld` if this isn't a codec that can decode bytes into something that can contain _links_; if it's a codec that doesn't involve links then it's a `serialization`, but given you're saying it's a public key then I suspect it's not even that.

I'd suggest it's probably just a `key`, but you also may want to consider the name `dms3-pub` to match the other public key types. But this also raises the question: is this actually a key type? Can you just use one (or more) of the existing key types for this?

yozgatsi:
so the intent is that it can decode bytes for new block types that will have links.
If you believe this is problematic or need more definition, then I can remove that now and keep only the multiaddr entry. we can revisit the former at a later time.

table.csv
rvagg: this objection still stands, can't be `ipld` here

yozgatsi:
Hi Rod,

A global, not just IPFS/LIBP2P specific, multicodec namespace registry can lead to conflict-free co-existence of multiple p2p networks in the wild.

The administrative role for such a registry would serve a function similar to the IANA.

If maintaining a global registry is outside the scope of this repository's mission, then say so and let's close this pull request.

I have a working p2p network that assigns the two codec values suggested above form the private-use range of values. My preference is to be transparent with all stakeholders by asking the maintainers of this repository to assign known codec values as discussed above.

Thanks
Tavit

PR #371, opened 2025-03-26
Update README.md

MasterJ93:
Added the Swift package in the Implementation section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The `validate.py` script can be used to validate the table once it's edited.
 - Zig
   - `multicodec` part of Zig project [multiformats-zig](https://github.com/zen-eth/multiformats-zig)
 - Swift
-  - `multicodec` part of [MultiformatsKit](https://github.com/ATProtoKit/MultiformatsKit)
+  - `Multicodec` part of [MultiformatsKit](https://github.com/ATProtoKit/MultiformatsKit)
 - [Add yours today!](https://github.com/multiformats/multicodec/edit/master/table.csv)
 
 ## Reserved Code Ranges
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,8 @@ The `validate.py` script can be used to validate the table once it's edited.
   - `multicodec` part of Kotlin project [multiformat](https://github.com/erwin-kok/multiformat)
 - Zig
   - `multicodec` part of Zig project [multiformats-zig](https://github.com/zen-eth/multiformats-zig)
+- Swift
+  - `multicodec` part of [MultiformatsKit](https://github.com/ATProtoKit/MultiformatsKit)
```
rvagg, line 88: Should this be `Multicodec` for Swift? since it's bundled inside the larger package this should be a pointer to users about how to access it.

MasterJ93: I don't understand. Are you asking me to point to the file that contains the code?

Sorry if I'm making things more difficult: I'm just trying to clarify what you'd like for me to do so I can resolve it.

rvagg:
```suggestion
  - `Multicodec` part of [MultiformatsKit](https://github.com/ATProtoKit/MultiformatsKit)
```

The use of `multicodec` like it is for the others isn't essential, don't just follow the same pattern if that's not idiomatic for your stack. I see lots of TitleCase in Swift and in your project in particular, so should it be `Multicodec` or perhaps something else?

MasterJ93: Sorry about that. I thought it was some kind of rule since everyone else was doing it. But yes: `Multicodec` is indeed correct here. I've resolved that by doing what you've suggested.

vmx (approved):

PR #383, opened 2025-06-20
Add transport-filecoin-piece-http

rvagg:
Intended for IPNI registrations: provider X has piece Y available on `transport-filecoin-piece-http`

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -149,6 +149,7 @@ car-multihash-index-sorted,     serialization,  0x0401,         draft,      CARv
 transport-bitswap,              transport,      0x0900,         draft,      Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,      Filecoin graphsync datatransfer
 transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP IPFS Gateway trustless datatransfer
+transport-filecoin-piece-http,  transport,      0x0930,         draft,      HTTP piece retrieval from Filecoin storage provider (typically /piece/CID endpoint)
 multidid,                       multiformat,    0x0d1d,         draft,      Compact encoding for Decentralized Identifers
 fr32-sha256-trunc254-padbintree,multihash,      0x1011,         draft,      A balanced binary tree hash used in Filecoin Piece Commitments as described in FRC-0069
 sha2-256-trunc254-padded,       multihash,      0x1012,         permanent,  SHA2-256 with the two most significant bits from the last byte zeroed (as via a mask with 0b00111111) - used for proving trees as in Filecoin
--- a/table.csv
+++ b/table.csv
@@ -149,7 +149,7 @@ car-multihash-index-sorted,     serialization,  0x0401,         draft,      CARv
 transport-bitswap,              transport,      0x0900,         draft,      Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,      Filecoin graphsync datatransfer
 transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP IPFS Gateway trustless datatransfer
-transport-filecoin-piece-http,  transport,      0x0930,         draft,      HTTP piece retrieval from Filecoin storage provider (typically /piece/CID endpoint)
+transport-filecoin-piece-http,  transport,      0x0930,         draft,      HTTP piece retrieval from Filecoin storage provider; https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0066.md
 multidid,                       multiformat,    0x0d1d,         draft,      Compact encoding for Decentralized Identifers
 fr32-sha256-trunc254-padbintree,multihash,      0x1011,         draft,      A balanced binary tree hash used in Filecoin Piece Commitments as described in FRC-0069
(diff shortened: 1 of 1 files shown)


masih (approved):

bajtos (approved):

rvagg:
> At the moment it seems like the same thing with a different endpoint

True, but we're reflecting existing realities here. Pieces are served exclusively via `/piece/` which has its own protocol rules (relatively simple), inner IPLD blocks within a piece are served via `/ipfs/` which have very different protocol rules. Byte range retrievals are probably the most important difference here being byte range requests for `raw` blocks. I suppose we could make sure that `dat-scope=block` supports standard http byte range requests and then we'd have ~compatibility (albeit with the additional need for a query string). But for now, you can at least use `/piece/` as a standard HTTP retrieval endpoint for large blobs, so you can, for example, point a video player at it and let it do its own byte range retrievals.

Perhaps we could work toward a convergence so it doesn't matter? But at that point, `transport-filecoin-piece-http` could mean that the piece is available on both endpoints, whereas `transport-ipfs-gateway-http` just means it's available on `/ipfs/`.

hsanjuan:
Is /piece retrieval ("protocol rules") specified anywhere? (I cannot readily find anything other than mentions in the `boost` docs).

I don't understand the byte-range support thing... does /piece retrieval allow byte ranges but then the the data cannot be verified after download even though it was requested by CID ? Is that a usecase?

masih:
[Here](https://github.com/filecoin-project/FIPs/blob/2c4e07a4c89cb6b04ce4815f46e4400fa273166a/FRCs/frc-0066.md?plain=1#L18) is the FRC

@rvagg please feel free to link it in the description of coded if it makes sense.

If there is anything else that's blocking this PR from being landed, please state explicitly. Otherwise, let's keep things moving.

aschmahmann (approved):
In the interest of not keeping anyone who is antsy here about how long approval will take I've approved the PR so @rvagg can make any changes and merge it at convenience.

A few things of note:

> [Here](https://github.com/filecoin-project/FIPs/blob/2c4e07a4c89cb6b04ce4815f46e4400fa273166a/FRCs/frc-0066.md?plain=1#L18) is the FRC

Is that the relevant spec, or is it really a modification to the [IPNI metadata table](https://github.com/ipni/specs/blob/5fffe5d807aa5b93d6e728dbfe6d4cc78224f780/IPNI.md#metadata) which would itself reference the FRC? My understanding is that nothing other than IPNI metadata uses these codes. Someone (perhaps @masih or @rvagg ) probably should update that metadata doc to describe what the expected the parameters for this code are and then @rvagg could include a link to it and/or FRC-66 in the table.

> I suppose we could make sure that dat-scope=block supports standard http byte range requests and then we'd have ~compatibility (albeit with the additional need for a query string)

TLDR: If using the safe piece CIDs then only the path change is required.

It's not `dag-scope=block` that you'd want since that still a CAR response (it returns the path blocks + the one at the end). It's `format=block` (or the [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) Accept header), or honestly in the case of Filecoin pieces addressed using the [Filecoin Piece multihash](https://github.com/filecoin-project/FIPs/blob/2c4e07a4c89cb6b04ce4815f46e4400fa273166a/FRCs/frc-0069.md#raw-codec) the raw codec will end up being used and no `format=block` is even necessary. i.e. a straight path change from `/piece` to `/ipfs` would do the job when using the safe piece CIDs.

masih:
There's no antsy-ness here. Filecoin frc is the spec. 

What this pr does is that it specifies a code to identify the transport. It so happens that IPNI is the only user of that identifier today. That code can be used anywhere inline with multicodec philosophy.

rvagg:
> safe piece CIDs

You're talking about [v2](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0069.md) I believe.

:ack: re `format=block` and `raw` etc., it's been too long. The other two concerns that make this still useful to identify a separate protocol/endpoint is (a) byte ranges and (b) the optionality of an SP _not_ serving /ipfs/ if they don't need to be doing so. Another future extension here is merkle proofs for arbitrary byte ranges also being served via /piece/ (somehow, TBD). I still quite like the model you mapped out for doing so with small piece ranges via /ipfs/ and maybe we end up doing that, but there's also a desire to ship something that's more transparently suitable for standard HTTP clients and tools (where merkle proofs might be optional), so /piece/ might be the best place to iterate on that. But maybe we just end up with both, or we iterate /ipfs/ to be all we need, :shrug:. I don't think using this entry rules out changing its meaning down the road.

table.csv
```diff
@@ -149,6 +149,7 @@ car-multihash-index-sorted,     serialization,  0x0401,         draft,      CARv
 transport-bitswap,              transport,      0x0900,         draft,      Bitswap datatransfer
 transport-graphsync-filecoinv1, transport,      0x0910,         draft,      Filecoin graphsync datatransfer
 transport-ipfs-gateway-http,    transport,      0x0920,         draft,      HTTP IPFS Gateway trustless datatransfer
+transport-filecoin-piece-http,  transport,      0x0930,         draft,      HTTP piece retrieval from Filecoin storage provider (typically /piece/CID endpoint)
```
rvagg, line 152:
```suggestion
transport-filecoin-piece-http,  transport,      0x0930,         draft,      HTTP piece retrieval from Filecoin storage provider; https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0066.md
```

PR #385, opened 2025-08-12
Add multicodec for DER-encoded X.509 certificate

joaopenteado:
This PR proposes a new multicodec entry for the canonical binary representation for X.509 certificates, as defined in [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1), encoded in DER form.

### Notes

- This covers exactly one Certificate object; if there’s demand for transporting multiple certs, CRLs, OCSP responses, or PKCS#7 bundles, those should be proposed as separate, unambiguous entries.
- Version is not restricted to v3 — v1 and v2 certificates are valid encodings and will parse under the same ASN.1 type.
- Consumers can unambiguously decode and parse the payload knowing only the multicodec code.


rvagg (approved):
sgtm

Issue #389, opened 2025-09-18
A few codecs are missing their counterparts (pub/priv keys)


vdods:
The following public key codecs don't have a private key counterpart:
- ed448-pub
- x448-pub
- mlkem-512-pub
- mlkem-768-pub
- mlkem-1024-pub
- jwk_jcs-pub

I see that the process to add codecs for the -priv versions of these would be to create a PR -- just wondering how to ensure I don't propose codec values that would collide with other PRs?

Thanks!

Apropos: the ed448-pub and x448-pub codecs were added here, but not the corresponding -priv codecs: https://github.com/multiformats/multicodec/pull/190

PR #390, opened 2025-09-18
Added missing codecs for the private key counterparts to a number of public key codecs.

vdods:
In particular, added:
- ed448-priv
- x448-priv
- mlkem-512-priv
- mlkem-768-priv
- mlkem-1024-priv
- jwk_jcs-priv

This is meant to address https://github.com/multiformats/multicodec/issues/389

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -197,6 +197,12 @@ bls12_381-g2-pub-share,         key,            0x130d,         draft,      BLS1
 bls12_381-g1-priv-share,        key,            0x130e,         draft,      BLS12-381 G1 private key share
 bls12_381-g2-priv-share,        key,            0x130f,         draft,      BLS12-381 G2 private key share
 sm2-priv,                       key,            0x1310,         draft,      SM2 private key
+ed448-priv,                     key,            0x1311,         draft,      Ed448 private key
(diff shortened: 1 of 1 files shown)


vmx:
What is your use case for the private keys?

Background: we try to add to the table only things that people actually use, so that we don't bloat it too much. There is also a discussion about private keys at e.g. https://github.com/multiformats/multicodec/pull/194. Making it harder for private keys to be shared, maybe makes it less likely that they accidentally leak. I think that's one of the reasons, why most of the keys currently don't have the `-priv` equivalent.

rvagg:
> It seems quite incomplete

Back to the original point @vmx is making - we don't just add things in here for completeness or else it'll be an unmaintainable, massive index of _things_. For instance, we've had requests to add basically all known mime types to the table, just for the sake of completeness. So we like to at least have some plausibility that there's a use-case connected to the additions and that they're not just to satisfy someone's desire for it to _feel_ more complete than what it is.

tbh this still sounds to me like you'd like to satisfy a feeling of symmetry in the table rather than having a concrete proposal for use. But you're welcome to outline more of a case for utility.

PR #392, opened 2025-11-21
Add ML-DSA public key multicodecs (FIPS 204)

vanderbr:
Add ML-DSA public key multicodecs (FIPS 204)

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -178,6 +178,9 @@ provenance-log-script,          serialization,  0x120a,         draft,      Veri
 mlkem-512-pub,                  key,            0x120b,         draft,      ML-KEM 512 public key; as specified by FIPS 203
 mlkem-768-pub,                  key,            0x120c,         draft,      ML-KEM 768 public key; as specified by FIPS 203
 mlkem-1024-pub,                 key,            0x120d,         draft,      ML-KEM 1024 public key; as specified by FIPS 203
+mldsa-44-pub,                   key,            0x1210,         draft,      ML-DSA 44 public key; as specified by FIPS 204
+mldsa-65-pub,                   key,            0x1211,         draft,      ML-DSA 65 public key; as specified by FIPS 204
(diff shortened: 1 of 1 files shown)


vmx (commented):
Why the gap and not directly after the ML-KEM keys starting with 0x120e?

vanderbr:
Because ML-KEM and ML-DSA are different post-quantum algorithms. ML-KEM (Kyber) is for key-establishment, while ML-DSA (Dilithium) is for digital signatures. NIST assigns them separate key types, so ML-DSA keys are not the same as  ML-KEM key types.   See https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf vs https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.203.pdf.  They are separate and multicodec should support both .  NIST has finalized ML-DSA as the standard signature scheme for post-quantum use, so any multicodec ecosystem that intends to support standardized PQ keys will eventually need an ML-DSA code point.  ML-KEM and ML-DSA keys are not interchangeable, and tooling that already supports ML-KEM will not be able to represent ML-DSA keys without a separate identifier. This mirrors how multicodec already distinguishes classical algorithms (RSA, Ed25519, ECDSA, etc.).

vanderbr:
Sorry, I didn't answer you question directly about why the gap ....  since ML-DSA is a different primitive than ML-KEM (signatures vs. key-establishment), giving it its own small block avoids collisions with future ML-KEM variants and keeps room for additional ML-DSA key types if NIST expands the line. Leaving the gap preserves forward-compatibility and follows existing multicodec practice where unrelated primitives are not forced to be sequential.

rvagg (approved):
fine by me, but @vanderbr note that we have `mlkem-*-priv` entries too, you might want to consider whether we should also have `mldsa-*-priv` block as well, happy to entertain that in a follow-up PR if you want it

PR #398, opened 2026-03-24
add entries for BIP-340: public/private key and varsig

MichaelMure:
While bip340 operates on secp256k1, it makes sense to have separate values:
- public keys have a different format: 32-byte x-only (bip340) vs 33-byte prefix+x (secp256k1)
- private keys have the same format but different usage expectation
- algorithm are different: schnorr (bip340) vs ECDSA (secp256k1)
- signatures have a different format: 64-byte [R.x, s] (bip340) vs 64-byte [r, s] (secp256k1)

That is, they can't reasonable mix or interoperate.

I tried to pick values that makes sense, let me know if I should change them.


dovydas55 (approved):

vmx (approved):
I haven't had a close look. But it looks reasonable. If no one else finds the time for a review, I think it's good to go.

PR #394, opened 2026-01-27
Create entries for SLH-DSA public keys

Wind4Greg:
This PR adds SLH-DSA public keys from [FIPS 205](https://csrc.nist.gov/pubs/fips/205/final).

I tried to do this in the same manner as the ML-DSA keys added in PR https://github.com/multiformats/multicodec/pull/392.

The keys are for the different SLH-DSA variants are enumerated in Table 2 section 11 of [FIPS 205](https://csrc.nist.gov/pubs/fips/205/final) . There are twelve variants:

SLH-DSA-SHA2-128s, SLH-DSA-SHAKE-128s, SLH-DSA-SHA2-128f, SLH-DSA-SHAKE-128f, SLH-DSA-SHA2-192s, SLH-DSA-SHAKE-192s, SLH-DSA-SHA2-192f, SLH-DSA-SHAKE-192f, SLH-DSA-SHA2-256s, SLH-DSA-SHAKE-256s, SLH-DSA-SHA2-256f, SLH-DSA-SHAKE-256f

I picked a contiguous block of what seem to be unassigned codes. BR Greg

PR #401, opened 2026-04-15
Add codes for SLH-DSA private keys

Wind4Greg:
Added codes for SLH-DSA private keys. Note that there are many flavors of SLH-DSA but only one type of private key for each. Took code values after the ML-DSA private key codes used in https://github.com/multiformats/multicodec/pull/399/changes#diff-bf5b449ed8c1850371f42808a186b5c5089edd0025700505a6b8f426cd54a6e4R224-R226.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -218,6 +218,18 @@ mlkem-512-priv,                 key,            0x1313,         draft,      ML-K
 mlkem-768-priv,                 key,            0x1314,         draft,      ML-KEM 768 public key; as specified by FIPS 203
 mlkem-1024-priv,                key,            0x1315,         draft,      ML-KEM 1024 public key; as specified by FIPS 203
 jwk_jcs-priv,                   key,            0x1316,         draft,      JSON object containing only the required members of a JWK (RFC 7518 and RFC 7517) representing the private key. Serialisation based on JCS (RFC 8785)
+slhdsa-sha2-128s-priv,          key,            0x131d,         draft,      SLH-DSA-SHA2-128s private key; as specified by FIPS 205
+slhdsa-shake-128s-priv,         key,            0x131e,         draft,      SLH-DSA-SHAKE-128s private key; as specified by FIPS 205
+slhdsa-sha2-128f-priv,          key,            0x131f,         draft,      SLH-DSA-SHA2-128f private key; as specified by FIPS 205
+slhdsa-shake-128f-priv,         key,            0x1320,         draft,      SLH-DSA-SHAKE-128f private key; as specified by FIPS 205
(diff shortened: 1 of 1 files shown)


msporny (approved):

dlongley (approved):

vmx (approved):
I first though, what are those for and why so many. Then I saw that it relates to https://github.com/multiformats/multicodec/pull/394 where the public keys were already added. There it was discussed that the private ones might be added later. That seems to be the case now.

Wind4Greg:
Thanks @vmx. SLH-DSA has three security levels, optimization  for  speed or size, and choice of SHA-2 or SHA-3  hash suites. Yikes. For initial credential use we are security category 1, optimized for size, using  SHA-2 suite. Cheers

lidel (approved):
LGTM given https://github.com/multiformats/multicodec/pull/394 landed in Jan makes sense.  I feel we will end up with a lot of dead codes that nobody uses, but hard to say which ones will end up being used, better to land and unblock implementations to figure it out.

@vmx coordination note: `0x131d`-`0x1328` sits directly after #399's    `0x1317`-`0x131c` so land #399 first and then rebase this one?

--- 

Out of curiosity looked into FIPS 205 and it defines 2 x 3 x 2 matrix = 12 parameter sets:

- **Hash function (2):** `sha2` or `shake`. SHA-2 is faster on CPUs with SHA-NI; SHAKE is preferred where Keccak is already in use (e.g., Ethereum(?), ML-KEM/DSA stacks).
- **Security level (3):** `128` / `192` / `256`. NIST security categories 1 / 3 / 5, matching AES-128/192/256.
- **Tradeoff (2):** `s` (small) or `f` (fast). Same security level, different Merkle-tree shape. `s` gives smaller signatures but slower signing; `f` gives faster signing but larger signatures. Rough ratio at 128-bit: ~7.8 KB / 580 ms vs ~17 KB / 30 ms (orders of magnitude on signing speed, ~2x on signature size, real world difference).

So each of the 12 is a legitimately different on-wire format with different performance and different signature bytes. Implementers need to pick one and stick with it, and the codec has to name which one.

When this PR is merged we will have complete coverage for all FIPS 205  variants with naming pattern `slhdsa-{sha2,shake}-{128,192,256}{s,f}-{pub,priv}` 👍

vmx:
https://github.com/multiformats/multicodec/pull/399 was merged, so this one can be rebased as @lidel suggests.

Wind4Greg:
Hi @vmx should I "rebase" this or  were you going to? I need to put some of the SLH-DSA private key prefixes into a W3C spec. Cheers Greg

PR #404, opened 2026-05-05
Add `radicle` namespace

lorenzleutgeb:
[Radicle](https://radicle.dev) is an open source, peer-to-peer code collaboration stack built on Git.

Repositories are addressed by hashing their initial "Repository Identity Document".

In order to allow addressing repositories by CIDs, reserve a new code.

For example, we currently use rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5 to address the repository which you can view on the web at https://radicle.network/nodes/seed.radicle.dev/rad%3Az3gqcJUoA1n9HaHKufZs5FCSGazv5.
The identifier here is what we call a "Repository Identifier" and it is a plain base56btc encoding of a special Git object in that repository.

With the imminent introduction of SHA-256 as the default object format/hash function in Git 3, we plan to introduce more flexibility into our repository identifier scheme. CIDs with this multicodec code look very promising.

### DIFF
--- a/table.csv
+++ b/table.csv
@@ -247,6 +247,7 @@ aes-gcm-256,                    encryption,     0x2000,         draft,      AES 
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
 sha256a,                        hash,           0x7012,         draft,      The sum of multiple sha2-256 hashes; as specified by Ceramic CIP-124.
+radicle,                        namespace,      0x8776,         draft,      Radicle namespace
 chacha20-poly1305,              multikey,       0xa000,         draft,      ChaCha20_Poly1305 encryption scheme
 blake2b-8,                      multihash,      0xb201,         draft,      Blake2b consists of 64 output lengths that give different hashes
 blake2b-16,                     multihash,      0xb202,         draft,


2color:
Since the hash that we intend to encode in the CID is of a Git object, I wonder if it makes sense for us to use `raw-git`: https://github.com/multiformats/multicodec/blob/45c88b89ab909c0fac7c86dafe43ad72d1e8e8a9/table.csv#L48

lorenzleutgeb:
I am not sure in how far semantics matter in this case. Sure, both things seem to capture SHA-1 and SHA-256 hashes, with the intention of identifying a Git object. But what about the intent/semantics of wanting to address a repository? This differs from the intent to address an object. Maybe someone can weigh in on how to balance such semantic differences.

For example, I have relatively good knowledge about what it would mean to address a Radicle repository (this currently happens, just using different syntax). But TBH I have little clue about how the `git-raw` codec is to be interpreted.

vmx:
This PR reminded me of the discussion at https://github.com/multiformats/multicodec/pull/203 about adding a custom code, although things are really Git.

In regards to what IPLD codecs are, see my comment at https://github.com/multiformats/multicodec/issues/204#issuecomment-765302334.

2color:
Thanks @vmx for the pointers. The consensus, based on my reading of #203 and #204, is that the multicodec describes how to decode bytes, not what they semantically address. 

A Radicle Repository Identifier is currently base58btc(Git commit object ID of the identity payload). The bytes that get hashed are a standard Git blob. The JSON inside carries the Radicle-specific structure. No new binary format. 
Screenshot 2026-05-07 at 12 51 06 pm

Alas, `git-raw` alone may not actually be sufficient to decode the Radicle identity document, but you could make that argument about other Git blob too: be it JSON, Rust source, a PNG, a Markdown file. `git-raw` is intentionally agnostic to that. What's more, there's a `manifest` file (`{"typeName":"xyz.radicle.id","version":1}`) in the commit of the Radicle Repository Identifier.

table.csv
```diff
@@ -247,6 +247,7 @@ aes-gcm-256,                    encryption,     0x2000,         draft,      AES
 silverpine,                     multiaddr,      0x3f42,         draft,      Experimental QUIC over yggdrasil and ironwood routing protocol
 sm3-256,                        multihash,      0x534d,         draft,
 sha256a,                        hash,           0x7012,         draft,      The sum of multiple sha2-256 hashes; as specified by Ceramic CIP-124.
+radicle-git,                    namespace,      0x8776,         draft,      Radicle Git Repository Identifier
```
lidel: This PR feels like a good opportunity to maybe write down policy in README? 

I agree  with prior art, for the *bytes* a RID addresses (a Git blob from `git hash-object`), `git-raw` (`0x78`, permanent) already decodes them. Per #204 that is what the CID codec field describes, so a Radicle-specific *data* codec seems unnecessary(?) as it duplicates `git-raw`.

iiuc what Radicle attempts to add here is a namespace, which feels fine and less controversial, if its framed as "namespace where git and other identifiers are used". 

`ipfs`, `ipns`, `swarm`, already  work this way: the namespace tags the system, the inner CID picks its own codec (for example, `ipfs` namespace uses CIDs with `dag-pb`, `raw`, `dag-cbor` and other byte codecs  internally).  


 `/radicle/<cid>` is the same shape, with the inner CID being `cid(git-raw, multihash(<algo>, <hash>))`. The Git 3 transition then needs no new codec: RIDs become `cid(git-raw, multihash(sha2-256, ...))` and the `radicle` namespace stays put. The hash algorithm is signaled by the inner multihash, not the namespace.

So maybe just clarify that `radicle` namespace proposed in this PR is the RIP-2? (or similar, having some spec document referenced removes ambiguity):

```suggestion
radicle,                        namespace,      0x8776,         draft,      Radicle namespace; payload is a Git object (git-raw 0x78), see RIP-2
```

ps. Would be good to link to [RIP-2](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad%3Az3trNYnLWS11cJWC6BbxDs5niGo82/tree/0002-identity.md) in the PR body but other than that feels sensible + Radicle been around for a long time.

ps2. for unconvinced, prior art precedent: `ipfs`, `ipns`, `swarm`, `streamid`, `lbry`, plus recent merges `adnl` (#402), Massa (#396), `shelter-*` (#369).

2color: Sorry about the naive question, but what are namespaces actually for? Feel free to link to relevant resources

lorenzleutgeb: > The Git 3 transition then needs no new codec: RIDs become `cid(git-raw, multihash(sha2-256, ...))`

This is understood, yes.

> So maybe just clarify that `radicle` namespace proposed in this PR is the RIP-2? (or similar, [...]

Similar. We would likely have a new RIP that amends RIP 2. RIP 2 talks about the "20 plain bytes" encoding. The question is what comes first: The RIP or the entry in this table? Since this table seems to allow "draft" entries, I figured that it might be acceptable to add to the table first, then write the RIP which can safely mention the codec that is allocated, and in another PR update the line to refer to that shiny new RIP. If Radicle does not deliver the RIP and the codec sees no adoption, since the entry is not "permanent", it could also be removed.

Regarding your suggestion: Radicle might in the future work on other VCSes, not only Git. With the concept of a namespace, it seems that we would have more control. In that case I would drop the talk about Git and `git-raw` from the listing, and instead define the rest in the RIP. Or go the other way and name the namespace `radicle-git`.

You can see that I do not really know what these namespaces mean, so I would also appreciate guidance here.

lidel: No worries, I am also tainted by too much history, so my understanding could be outdated, but here it is: a `namespace` tags *which system* an identifier belongs to, used either as a path prefix or as a binary prefix in wire formats. IIRC there was a time at Protocol Labs where we wanted to have Plan9-like future, where all systems/protocols lived under one `/`, so  the meta-namespace is `path` (`0x2f`, the `/` byte) at `table.csv:29`; and then  specific system namespaces (`ipfs`, `ipns` etc) nest under it. 

Concrete usage:

- **Paths** (mostly IPFS, mostly historical): `/ipfs/<cid>/foo`, `/ipns/k51.../foo`, `/swarm/<addr>/...`. Intention was to unify gateway URLs, routing keys, provider records etc.
- **ENS `contenthash`** (iiuc the main user of "namespaces" today?  ([ERC-1577](https://eips.ethereum.org/EIPS/eip-1577), [ENSIP-7](https://docs.ens.domains/ensip/7/)): binary form `<protoCode uvarint><CIDv1>`, where `protoCode` is `ipfs-ns` (`0xe3`), `ipns-ns` (`0xe5`), or `swarm-ns` (`0xe4`). For example, an ENS record that resolves to IPFS encodes as `0xe3 || 0x01 0x70 0x12 0x20 <32-byte hash>` (ipfs namespace + CIDv1 of `dag-pb` + `sha2-256`). The namespace byte tells the resolver which network to dispatch to; the inner CID picks its own codec.


As  for more historical references,  there is old https://github.com/multiformats/multiformats/pull/55, which aimed to formalize `<namespace-varint><body>` as the wire pattern.

Loosely quoting  Stebalien on this exact point in https://github.com/multiformats/multicodec/issues/204#issuecomment-765811000:

> Within [system], use CIDs. Outside of [system], use namespaced paths, or namespaced CIDs if you need something shorter. E.g., an ENS record might refer to `<ipfs-codec><cidv1><...>` and/or `/ipfs/CIDv1/...`

So the `radicle` namespace slots into the same machinery: ENS-style records, HTTP gateway URLs, and multipath routers can dispatch on the prefix without parsing the inner CID. 

(But in practice, I only seen `namespace` codes being used in ENS/Ethereum and others using `contenthash` contracts)

lorenzleutgeb: Okay, so instead of using
 - `<cid(radicle-git, multihash(sha2-256, ...))>`
 - `<cid(radicle-git, multihash(sha1, ...))>`
 - `<cid(radicle-hg, multihash(sha1, ...))>` (this one is fictional)

we would use

 - `<radicle-codec><cid(git-raw, multihash(sha2-256, ...))>`
 - `<radicle-codec><cid(git-raw, multihash(sha1, ...))>`
 - `<radicle-codec><cid(hg-raw, multihash(sha1, ...))>` (this one is fictional)

That sounds fine to me. How would that entry look like then?

```suggestion
radicle,                        namespace,      0x8776,         draft,      Radicle namespace
```

lorenzleutgeb:
Updated to the effect of my suggestion in https://github.com/multiformats/multicodec/pull/404#discussion_r3204041724 after that got a thumbs up by @lidel.

table.csv
aschmahmann: > (But in practice, I only seen namespace codes being used in ENS/Ethereum and others using contenthash contracts)

I have no specific objection to adding a radicle namespace, although from this discussion it doesn't seem like there's an intended use for the namespace at the moment.

Currently we see namespace requests exclusively for the use case documented above (which IIUC isn't something Radicle is looking for at the moment). However, it certainly seems reasonable to use `<namespace><system-address>` (or in Radicle's case `<namespace><cid>`) like ENS does to allow compact representation of addresses across multiple systems.

If there's an interest in getting the namespace code now then SGTM, if however it's not known how Radicle would use the namespace yet (or when it would be used instead of just a CID) it might make sense for the authors to park this until the need arises.

lorenzleutgeb: We would use `<radicle-codec><cid(git-raw, multihash(sha1, ...))>` ASAP for repository identifiers.

E.g. https://radicle.network/nodes/seed.radicle.dev/rad%3Az3gqcJUoA1n9HaHKufZs5FCSGazv5/ would also be reachable via `https://radicle.network/nodes/seed.radicle.dev/rad%3A<radicle-codec><cid(git-raw, multihash(sha1, 039fb5bb70789bed3e1f026ddb0c417ded37c3365882))>`.

Is that somehow unreasonable?

lidel: > We would likely have a new RIP that amends RIP 2. [..]
> The question is what comes first: The RIP or the entry in this table?

Sounds reasonable, no objection to the wire shape. Older entries like `lbry` (#332) were merged with a one-liner linking to project's spec, so the bar is low, we are just trying to raise it a bit :)   Goal: someone reading `table.csv`'s git history in five years can trace `radicle 0x8776` back to a specifying document. 

Since the bar is low: just link to the new RIP once you have a draft, and we can merge this.

2color: A follow-up we'll need to settle is the stringified form. CID's [stringified form multibase-prefixes only the CID](https://github.com/multiformats/cid#variant---stringified-form). 

For the namespace there seem to be two precedents:
- path/name form (/radicle/<cid-string>, like /ipfs/…) 
- ENS-style binary <namespace-varint><CIDv1>
 
Within our own `rad:` URI scheme the `radicle` namespace is arguably redundant (the scheme already names the system), but there's maybe a case to be made to also include the namespace for consistency?

Does that match how you'd expect the namespace to be used?

In other words, if we continue to use Base58 for RIDs and want to include the namespace codec in it, is it considered valid to `multibase(base58, <namespace><CID>)`? (update: it is evidently not a valid CID)

Based on https://github.com/multiformats/multicodec/issues/204#issuecomment-765811000, my sense is that the namespace codec buys us little within Radicle: we have our own URI scheme and use it whenever we pass RIDs around, so internally a bare CID suffices. The namespace would only earn its keep for external interop (ENS-style records, multipath resolvers), and since namespace codecs seem narrowly used in practice, I'm struggling to see how much value we'd get from it.

lidel: Your read is right, and I think the two "precedents" you listed aren't competing options. They're the text form and the binary form of the same namespace, and they map cleanly onto two things Radicle already has or could have.

> [!NOTE]
> I may be over-correcting towards pragmatism below, maybe there is more formal nuance in historical threads, but i'm not sure if formal rigor survived even the before-nucleation test of time.

[In practice, when used in real world] A `namespace` code is just the compact binary equivalent of a URI scheme.  `rad:` and `0x8776` encode the same fact, "this identifier belongs to Radicle, use Radicle-specific tooling to resolve it if it matters", one as ASCII, one as a varint. They convert losslessly and you pick by context:

- **Text contexts** (URLs, configs, copy-paste): the scheme, `rad:<cid>` or `/radicle/<cid>`. Here the inner thing stays a bare, valid CID and every CID tool keeps working. This is what `rad:` already is; `rad:` is your `/radicle/` prefix.
- **Binary contexts where text is wasteful or infeasible** (ENS-style `contenthash` records, multipath resolvers, fixed-width fields): `<0x8776><CIDv1>`, optionally multibase-wrapped for transport.

The same duality already exists for IPFS, in how some systems store IPFS CIDs: `ipfs://<cid>` and `/ipfs/<cid>` in text, `0xe3<CIDv1>` in ENS `contenthash`. ENS chose the binary form because on-chain bytes cost gas, which is the whole point of the namespace code: it earns its keep wherever a plain-text URI is too big or can't be stored.

On your CID question: correct, `multibase(base58, <0x8776><CID>)` is not a CID, and that's expected. The namespace prefix produces a namespaced byte string that a namespace-aware parser unwraps before parsing the inner CID; a plain CID parser will reject it. So the two designs are: keep `rad:` as the scheme and pass bare CIDs (stays a valid CID), or emit `<namespace><CID>` as one opaque binary token (compact, not a CID). Don't mix them, since putting the varint inside a `rad:` URI is redundant and breaks CID validity, exactly as you found.

So the value question is genuinely Radicle's to make, not multicodec's to gatekeep. Within Radicle, you have `rad:` and a bare CID suffices. The `0x8776` code only matters if you foresee a compact cross-system binary identifier (an ENS record, a multipath router, anything where a text URI is too costly). If you do, reserving it now is cheap and well-precedented (the whole `ipfs`/`ipns`/`swarm`/`streamid`/`lbry`/`massa`/`adnl` block was reserved on the same basis, and it's `draft`, so removable). If you don't, the `rad:` scheme already does the job and you can park this until the need is real.

lidel (approved):
@vmx if @lorenzleutgeb  finds this namespace still useful, lgtm as a draft (we already accepted way more experimental namespaces).

One thing worth saying out loud: whether you actually want this namespace code is your call, @lorenzleutgeb, not ours to push. As I noted [above](https://github.com/multiformats/multicodec/pull/404#discussion_r3427959468), if you're happy passing plain CIDs around inside `rad:` URIs, you may not need `0x8776` at all. It only really earns its keep for compact binary interop (ENS-style records, multipath resolvers, anywhere a text URI is too costly). Reserving it now is cheap and easy to drop later, so no harm either way. Totally fine to take it, or to park it until the need is real.

Tiny nit before merge: please point the description at your spec, so anyone reading the table later can trace `0x8776` or URIs back to a relevant-ish document:

```suggestion
radicle,                        namespace,      0x8776,         draft,      Radicle namespace; see RIP-2 or later
```

vmx (approved):
Thanks @lidel for diving into the details. I didn't know much about namespaces before. I agree that if @lorenzleutgeb still sees a use for this namespace I'm happy to see this merged (with the additional reference, so people in the future can find out more details what this is about).

PR #405, opened 2026-05-23
Add codecs for archive formats

### LINKED ISSUE #105: Request: Archive Formats
Examples: tar, zip

ofek:
Resolves https://github.com/multiformats/multicodec/issues/105


vmx:
You PR adds quite a lot of new entries. What is your use case for those? We try to grow the table as needed on specific uses. Historically there have been entries which "flooded" the table. We try to avoid that these days.

